PR #8, opened 2014-12-24
AddAlbum instead of PlayAlbum

erdoukki:
Enhancement of SqueezeboxQobuz.user.js
Queue album from Qobuz website instead of only playing the album.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -454,7 +454,7 @@ sub QobuzGenre {
 				type    => 'editor-picks',
 			}]
 		},{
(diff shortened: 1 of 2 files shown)


erdoukki:
I have tested your enhanced version of the addalbum.
Works great.
Thanks.

Itersting proposal to make optional the option to queue the tracks instead of directly play them.

May be a key modifier (control ?) will be sufficient.

mherger:
Ok, committed. Thanks!

PR #17, opened 2019-04-13
prefer track performer over album artist

marcelveldt:
Prefer the main performer of the track as artist instead of the album's artist to handle various artists albums.

As it is now, when listing playlists with tracks in it from various artists (compilation) albums, the artist is listed as "Various Artists" which is kind of annoying as the track itself actually has a performer.

This simple adjustment fixes that where the performer of the track is used instead of the album's artist.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -988,7 +988,7 @@ sub _playlistItem {
 sub _trackItem {
 	my ($client, $track, $isWeb) = @_;
 
-	my $artist = $track->{album}->{artist}->{name} || $track->{performer}->{name} || '';
+	my $artist = $track->{performer}->{name} || $track->{album}->{artist}->{name} || '';
 	my $album  = $track->{album}->{title} || '';
 
 	my $item = {


michaelherger:
Looking good to me. Would you have a playlist I can test this with?

marcelveldt:
https://open.qobuz.com/playlist/2208363

michaelherger:
Hmm... just had this case:

```
                                          {
                                            album                   => {
                                                                         artist                  => {
                                                                                                      albums_count => 66,
                                                                                                      id => 39_723,
                                                                                                      image => undef,
                                                                                                      name => "Taj Mahal",
                                                                                                      picture => undef,
                                                                                                      slug => "taj-mahal",
                                                                                                    },
                                                                                                    ....
                                                                       },
                                            performer               => { id => 2_126_398, name => "Various Interprets" },
                                            ...
                                          }
```

In this case your change would be wrong, right?

marcelveldt:
Yeah, it's an edge case. The Qobuz api is actually wrong (also in their own web interface).
The performer is set to "various interprets" but in fact it should be the "Main Artist" in the performers list but that one is really strange formatted.

michaelherger:
> https://open.qobuz.com/playlist/2208363

How would I be able to open this? It only takes me to the main page.

michaelherger:
Are you saying that the one I found was an edge case? Let's give it a try and see who's complaining. I'll bump the plugin later today. It's sunny right now. Need to do some stuff outside :-)

marcelveldt:
> Are you saying that the one I found was an edge case? Let's give it a try and see who's complaining. I'll bump the plugin later today. It's sunny right now. Need to do some stuff outside :-)

Yeah, i saw it myself too while testing. It can be fixed but then we need to parse the strings inside the perfomers. Not very good...

This is the piece of code I use in some python playlistsyncer I'm creating:

> for artist_str in track["performers"].split(" - "):
>                     if "Performer" in artist_str or "MainArtist" in artist_str or "Main Artist" in artist_str:
>                         artist = artist_str.split(", ")[0]
>                         if artist not in item["artists"]:
>                             item["artists"].append(artist)

marcelveldt:
btw: I think there's another adjustment needed too because it seems that there is also some precaching going on which does not share the same code. 

https://github.com/pierrepoulpe/SqueezeboxQobuz/blob/master/API.pm#L637

marcelveldt:
> Are you saying that the one I found was an edge case? Let's give it a try and see who's complaining. I'll bump the plugin later today. It's sunny right now. Need to do some stuff outside :-)

Have fun outside, no rush. I'll try to look into the edge case myself too but that that strong in Perl code ;-)

michaelherger:
https://github.com/pierrepoulpe/SqueezeboxQobuz/commit/ca043f851a7f376b6d6038eb4abc8cced4d7a7b2

marcelveldt:
Thanks!

PR #27, opened 2023-02-28
Add new display options.

SamInPgh:
Add options to 1) display the release year with Albums and 2) display Artist Albums sorted by release date, either ascending or descending, in addition to alphabetically.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1130,6 +1130,7 @@ sub _albumItem {
 	if ($albumName) {
 		$item->{line1} = $albumName;
 		$item->{line2} = $artist . ($albumYear ? ' (' . $albumYear . ')' : '');
+		$item->{name} .= $albumYear ? "\n(" . $albumYear . ')' : '';
 	}
 
 	if ( $album->{released_at} > time  || (!$album->{streamable} && !$prefs->get('playSamples')) ) {
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -467,18 +467,22 @@ sub QobuzArtist {
 		if ($artist->{albums}) {
 			my $albums = [];
 
-			my $sortByDate = (preferences('server')->get('jivealbumsort') || '') eq 'artflow';
+			# sort by release date if requested
+			my $sortByDate = $prefs->get('sortArtistAlbums');
 
 			$artist->{albums}->{items} = [ sort {
-
-				# push singles and EPs down the list
-				if ( ($a->{tracks_count} >= 4 && $b->{tracks_count} < 4) || ($a->{tracks_count} < 4 && $b->{tracks_count} >=4) ) {
-					return $b->{tracks_count} <=> $a->{tracks_count};
+				if ($sortByDate) {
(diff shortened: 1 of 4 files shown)


Plugin.pm
```diff
@@ -467,18 +467,22 @@ sub QobuzArtist {
...
+						return $b->{tracks_count} <=> $a->{tracks_count};
+					}
 
-				return $a->{released_at}*1 <=> $b->{released_at}*1 if $sortByDate;
-				return lc($a->{title}) cmp lc($b->{title});
-
+					return lc($a->{title}) cmp lc($b->{title});
+				}
+				
 			} @{$artist->{albums}->{items} || []} ];
```
michaelherger, line 486: That's certainly doing the job. 

If we were picky about performance, or dealing with large lists, you could optimize it: in its current state the condition for `$sortByDate` would be evaluated for each and every item. That's probably not much of a problem, as hardly any artist would have more than a few albums (besides Frank Zappa, who's got something between 100-200 on Spotify, IIRC...). But you could evaluate it outside the loop, then create an anonymous sub to execute the sorting comparison. Something like:


```perl
my $sorter;

if ($sortByDate) {
   if ($sortByDate == 1) {
      $sorter = sub { $b->{released_at}*1 <=> $a->{released_at} };
   } else {
      $sorter = sub { $a->{released_at}*1 <=> $b->{released_at} };
   }
}
else {
   $sorter = sub {
      if ( ($a->{tracks_count} >= 4 ...) {
         return $b->{tracks_count} <=> $a->{tracks_count};
      }

      return lc($a->{title})...;   
   };
}

$artist->{albums}->{items} = [ sort $sorter @{$artist->{albums}->{items} || []} ];
```

I guess you get the idea. But as I said: that's just an idea, not really needed here (nor fully implemented or tested).

Plugin.pm
```diff
@@ -1124,7 +1129,8 @@ sub _albumItem {
 
 	if ($albumName) {
 		$item->{line1} = $albumName;
-		$item->{line2} = $artist;
+		$item->{line2} = $artist . ($albumYear ? ' (' . $albumYear . ')' : '');
+		$item->{name} .= $albumYear ? "\n(" . $albumYear . ')' : '';
```
michaelherger, line 1133: Which UI would respect that line feed ("`\n`")? If it's always respected, I'd put the year without the brackets, as they look odd, alone on a single line:

```
Greatest Album  - Favorite Artist
(1999)
```

michaelherger (commented):
Good job! 

I left a comment "for educational purposes". And was wondering about the year in brackets. But if you like like that, I can certainly live with it.

Plugin.pm
SamInPgh: Good point. I'm usually big on code efficiency but this is my first experience with sorting in Perl and I admit there was some cutting and pasting going on here.  ;-)

Plugin.pm
SamInPgh: As far as I can tell, only the native LMS skins use the 'name' field to display the album information. All the other clients use 'line1' and 'line2'. As mentioned on the forum thread, I used the brackets on the year to be consistent with how the LMS skins display the release year with library albums when the Settings->Interface option to do so is specified.

PR #29, opened 2023-03-16
Prepend the name of the 'work' to track titles when present.

SamInPgh:
I feel like I might have done something wrong in the commit/pull process. Let me know if I did.

See discussion of issue here:  https://forums.slimdevices.com/forum/user-forums/3rd-party-software/94305-qobuz-com-streaming-plugin?p=1633997#post1633997


michaelherger:
This would only add the "Work" title in the Now Playing area, but eg. not when importing content to the library, right?

SamInPgh:
Correct.

PR #30, opened 2023-03-17
Populate the 'year' in track items and correct a typo.

SamInPgh:
Populating the 'year' is for consistency only. Currently it shows up as 'undefined' in the debug log, which can be confusing. The typo is more important and could allow tracks from as-yet-unreleased albums to be played, which Qobuz probably wouldn't like.  ;-)

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1212,7 +1212,11 @@ sub _trackItem {
 		$item->{work} = $track->{work};
 	}
 
-	if ($track->{released_at} && $track->{released_at} > time) {
+	if ( $track->{album} ) {
+		$item->{year} = $track->{album}->{year} || (localtime($track->{album}->{released_at}))[5] + 1900 || 0;
+	}
+
+	if ($track->{album} && $track->{album}->{released_at} && $track->{album}->{released_at} > time) {
 		$item->{items} = [{
 			name => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED'),
 			type => 'textarea'


michaelherger:
I think it's valid to check for a track's release date. Some tracks are released without an album, or months before the album is released. And such tracks should be playable.

SamInPgh:
From what I can see in the debug logs, there is no 'released_at' attribute defined for tracks --- only for albums. Therefore the 'if' would always evaluate as false.

SamInPgh:
Also, if a track were to be released on its own ahead of the album, as in your example, then $track->{album} would presumably be 'null' so it would still be playable according to the logic.

SamInPgh:
Looking into this a little further, it appears that Qobuz handles those advance-release "singles" by creating an album with the same name as the track and a track-count of 1 --- sort of like a mini EP. This album entry is where the 'released_at' attribute is stored. For example, Laure Del Ray has released 3 singles from her upcoming album over the last few months and each of them appears to the API as a one-track album. If a track item should ever appear without an associated album item, the 'released_at' value couldn't be checked and the track would be deemed playable, assuming it met the other criteria such as being streamable, etc.

michaelherger:
That makes sense. And it probably also is one of the reason why some favorites can disappear again: once a track has been released on album, the streaming services could pull the standalone versions again. I'll look into this as soon as I find time. Thanks!

Plugin.pm
```diff
@@ -1212,7 +1212,11 @@ sub _trackItem {
 		$item->{work} = $track->{work};
 	}
 
-	if ($track->{released_at} && $track->{released_at} > time) {
+	if ( $track->{album} ) {
+		$item->{year} = $track->{album}->{year} || (localtime($track->{album}->{released_at}))[5] + 1900 || 0;
```
michaelherger, line 1216: Haha... I wanted to comment that this would never be "0", as we'd always add 1900 to the result of `released_at`. But then I figured out I'm using the same logic all over the plugin...

PR #32, opened 2023-04-03
Classical music enhancements

darrell-k:
New plugin options to enable display of classical works/composers.

The display is controlled by the Qobuz tag genre_list: if this array contains the element "Classique" and the user has enabled "Enhanced classical music display" in the options, subheadings will be created for each composer/work in the release, and the titles of tracks added to the play queue/now playing will have the composer and work prepended.

There is also an option for the user to control the playlists that are created in the listing for each work: they can appear before the individual tracks (this is the default), after the individual tracks, or not at all. 

If the user finds a release which is not tagged "Classique" (see above), but would nonetheless benefit from these enhancements, there is an option for them to add the genre to a list in their plugin settings.

Finally, various string manipulation has been implemented to try to clean up the somewhat inconsistent composer/work/title tags returned by Qobuz.

Note that these affect Qobuz listings and the play queue/now playing displays only. There is no change to the data sent to the user's library by the online music import.

Many thanks to SamY for starting this work, and his testing and encouragement.

### DIFF
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -32,9 +32,19 @@ my $cache;
 my $prefs = preferences('plugin.qobuz');
 my $log = logger('plugin.qobuz');
 my $isClassique;
+my %genreList;
+
+initGenreMap();
+
+$prefs->setChange(\&initGenreMap, 'classicalGenres');
+
 sub init {
 	my $class = shift;
 	return pack('H*', $_[0]) =~ /^(\d{9})(.*)/
+}
+
+sub initGenreMap {
+   %genreList = map { $_ => 1 } split /\s*,\s*/, $prefs->get('classicalGenres');
 }
 
 sub getCache {
@@ -110,9 +120,8 @@ sub _precacheAlbum {
 		# If the user pref is for classical music enhancements to the display, is this a classical release or has the user added the genre to their custom classical list?
 		$isClassique = 0;
 		if ( $prefs->get('useClassicalEnhancements') ) {
-			my %genreList = map { $_ => 1 } split '\s*,\s*', $prefs->get('classicalGenres');	
 			if ( $album->{genres_list} && grep(/Classique/,@{$album->{genres_list}}) || $genreList{$album->{genre}} ) {
-			$isClassique = 1;
+				$isClassique = 1;
 			}
 		}
 
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -31,7 +31,7 @@ use constant QOBUZ_STREAMING_FLAC_HIRES2 => 27;
 my $cache;
 my $prefs = preferences('plugin.qobuz');
 my $log = logger('plugin.qobuz');
-
+my $isClassique;
 sub init {
 	my $class = shift;
 	return pack('H*', $_[0]) =~ /^(\d{9})(.*)/
@@ -107,6 +107,15 @@ sub _precacheAlbum {
 		$album->{genre} = $album->{genre}->{name};
 		$album->{image} = __PACKAGE__->getImageFromImagesHash($album->{image}) || '';
 
+		# If the user pref is for classical music enhancements to the display, is this a classical release or has the user added the genre to their custom classical list?
+		$isClassique = 0;
+		if ( $prefs->get('useClassicalEnhancements') ) {
+			my %genreList = map { $_ => 1 } split '\s*,\s*', $prefs->get('classicalGenres');	
+			if ( $album->{genres_list} && grep(/Classique/,@{$album->{genres_list}}) || $genreList{$album->{genre}} ) {
+			$isClassique = 1;
+			}
+		}
+
 		my $albumInfo = {
 			title  => $album->{title},
 			id     => $album->{id},
@@ -116,6 +125,7 @@ sub _precacheAlbum {
 			goodies=> $album->{goodies},
 			genre  => $album->{genre},
 			genres_list => $album->{genres_list},
+			isClassique => $isClassique,
 		};
 
 		_precacheTracks([ map {
@@ -155,7 +165,7 @@ sub precacheTrack {
 
 	my $album = $track->{album} || {};
 	$track->{composer} ||= $album->{composer} || {};
-
+	
 	my $meta = {
 		title    => $track->{title} || $track->{id},
 		album    => $album->{title} || '',
@@ -173,6 +183,7 @@ sub precacheTrack {
 		work     => $track->{work},
 		genre    => $album->{genre},
 		genres_list => $album->{genres_list},
+		isClassique => $isClassique,
 	};
 
 	if ($track->{audio_info} && defined $track->{audio_info}->{replaygain_track_gain}) {
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -95,7 +95,7 @@ sub _precacheAlbum {
 	$albums = __PACKAGE__->filterPlayables($albums);
 
 	foreach my $album (@$albums) {
-		foreach (qw(composer duration genres_list articles article_ids catchline
+		foreach (qw(composer duration articles article_ids catchline
 			# maximum_bit_depth maximum_channel_count maximum_sampling_rate maximum_technical_specifications
 			popularity previewable qobuz_id sampleable slug streamable_at subtitle created_at
 			product_type product_url purchasable purchasable_at relative_url release_date_download release_date_stream release_date_original
@@ -114,6 +114,8 @@ sub _precacheAlbum {
 			image  => $album->{image},
 			year   => (localtime($album->{released_at}))[5] + 1900,
 			goodies=> $album->{goodies},
+			genre  => $album->{genre},
+			genres_list => $album->{genres_list},
 		};
 
 		_precacheTracks([ map {
@@ -169,6 +171,8 @@ sub precacheTrack {
 		goodies  => $album->{goodies},
 		version  => $track->{version},
 		work     => $track->{work},
+		genre    => $album->{genre},
+		genres_list => $album->{genres_list},
 	};
 
 	if ($track->{audio_info} && defined $track->{audio_info}->{replaygain_track_gain}) {
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -120,7 +120,7 @@ sub _precacheAlbum {
 		# If the user pref is for classical music enhancements to the display, is this a classical release or has the user added the genre to their custom classical list?
 		$isClassique = 0;
 		if ( $prefs->get('useClassicalEnhancements') ) {
-			if ( $album->{genres_list} && grep(/Classique/,@{$album->{genres_list}}) || $genreList{$album->{genre}} ) {
+			if ( ( $album->{genres_list} && grep(/Classique/,@{$album->{genres_list}}) ) || $genreList{$album->{genre}} ) {
 				$isClassique = 1;
 			}
 		}
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -46,15 +46,12 @@
 				<option [% prefs.useClassicalEnhancements ? "selected" : "" %] value="1" >[% "YES" | string %]</option>
 				<option [% prefs.useClassicalEnhancements ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
 			</select>
-			<!--[% "PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" | string %]
-			[% "PLUGIN_QOBUZ_CLASSICAL_GENRES" | string %]&nbsp;
-			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">-->
 		[% END %]
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION" desc="" %]
 			<select name="pref_workPlaylistPosition" id="workPlaylistPosition">
-			  <option [% prefs.workPlaylistPosition == "before" ? "selected" : "" %] value="before" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_BEFORE" | string %]</option>
-			  <option [% prefs.workPlaylistPosition == "after"  ? "selected" : "" %] value="after" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER" | string %]</option>
-		          <option [% prefs.workPlaylistPosition == "hidden" ? "selected" : "" %] value="hidden" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN" | string %]</option>
+				<option [% prefs.workPlaylistPosition == "before" ? "selected" : "" %] value="before" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_BEFORE" | string %]</option>
+				<option [% prefs.workPlaylistPosition == "after"  ? "selected" : "" %] value="after" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER" | string %]</option>
+				<option [% prefs.workPlaylistPosition == "hidden" ? "selected" : "" %] value="hidden" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN" | string %]</option>
 			</select>
 		[% END %]
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_CLASSICAL_GENRES" desc="PLUGIN_QOBUZ_CLASSICAL_GENRES_DESC" %]
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -40,7 +40,7 @@
 				<option [% prefs.showYearWithAlbum ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
 			</select>
 		[% END %]
-		<hr style="opacity:0.4" class="sub-sep"/>
+		<hr class="sub-sep"/>
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS" desc="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" %]
 			<select name="pref_useClassicalEnhancements" id="useClassicalEnhancements">
 				<option [% prefs.useClassicalEnhancements ? "selected" : "" %] value="1" >[% "YES" | string %]</option>
@@ -58,7 +58,7 @@
 			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">
 		[% "PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_TEXT" | string %]
 		[% END %]
-		<hr style="opacity:0.4" class="sub-sep"/>
+		<hr class="sub-sep"/>
 
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_IMPORT" desc="" %]
 			<input type="checkbox" [% IF prefs.pref_dontImportPurchases %]checked [% END %] class="stdedit" name="pref_dontImportPurchases" id="pref_dontImportPurchases" value="1" />
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -41,6 +41,15 @@
 			</select>
 		[% END %]
 
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS" desc="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" %]
+			<select name="pref_useClassicalEnhancements" id="useClassicalEnhancements">
+				<option [% prefs.useClassicalEnhancements ? "selected" : "" %] value="1" >[% "YES" | string %]</option>
+				<option [% prefs.useClassicalEnhancements ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
+			</select>
+			&nbsp;[% "PLUGIN_QOBUZ_CLASSICAL_GENRES" | string %]&nbsp;
+			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">
+		[% END %]
+
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_IMPORT" desc="" %]
 			<input type="checkbox" [% IF prefs.pref_dontImportPurchases %]checked [% END %] class="stdedit" name="pref_dontImportPurchases" id="pref_dontImportPurchases" value="1" />
 			<label for="pref_dontImportPurchases" class="stdlabel">[% "PLUGIN_QOBUZ_DONT_IMPORT_PURCHASES" | string %]</label>
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -50,6 +50,13 @@
 			[% "PLUGIN_QOBUZ_CLASSICAL_GENRES" | string %]&nbsp;
 			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">-->
 		[% END %]
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION" desc="" %]
+			<select name="pref_workPlaylistPosition" id="workPlaylistPosition">
+			  <option [% prefs.workPlaylistPosition == "before" ? "selected" : "" %] value="before" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_BEFORE" | string %]</option>
+			  <option [% prefs.workPlaylistPosition == "after"  ? "selected" : "" %] value="after" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER" | string %]</option>
+		          <option [% prefs.workPlaylistPosition == "hidden" ? "selected" : "" %] value="hidden" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN" | string %]</option>
+			</select>
+		[% END %]
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_CLASSICAL_GENRES" desc="PLUGIN_QOBUZ_CLASSICAL_GENRES_DESC" %]
 			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">
 		[% "PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_TEXT" | string %]
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -40,15 +40,21 @@
 				<option [% prefs.showYearWithAlbum ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
 			</select>
 		[% END %]
-
+		<hr style="opacity:0.4" class="sub-sep"/>
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS" desc="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" %]
 			<select name="pref_useClassicalEnhancements" id="useClassicalEnhancements">
 				<option [% prefs.useClassicalEnhancements ? "selected" : "" %] value="1" >[% "YES" | string %]</option>
 				<option [% prefs.useClassicalEnhancements ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
 			</select>
-			&nbsp;[% "PLUGIN_QOBUZ_CLASSICAL_GENRES" | string %]&nbsp;
-			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">
+			<!--[% "PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" | string %]
+			[% "PLUGIN_QOBUZ_CLASSICAL_GENRES" | string %]&nbsp;
+			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">-->
 		[% END %]
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_CLASSICAL_GENRES" desc="PLUGIN_QOBUZ_CLASSICAL_GENRES_DESC" %]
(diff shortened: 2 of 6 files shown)


HTML/EN/plugins/Qobuz/settings/basic.html
```diff
@@ -40,6 +40,28 @@
...
 		[% END %]
+		<hr style="opacity:0.4" class="sub-sep"/>
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS" desc="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" %]
+			<select name="pref_useClassicalEnhancements" id="useClassicalEnhancements">
+				<option [% prefs.useClassicalEnhancements ? "selected" : "" %] value="1" >[% "YES" | string %]</option>
+				<option [% prefs.useClassicalEnhancements ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
+			</select>
+			<!--[% "PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" | string %]
+			[% "PLUGIN_QOBUZ_CLASSICAL_GENRES" | string %]&nbsp;
+			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">-->
```
michaelherger, line 51: Please remove unused code. You can still leave it in your clone by only un-staging these lines (easily done with something like SourceTree) if you want them to be around for future reference.

HTML/EN/plugins/Qobuz/settings/basic.html
```diff
@@ -40,6 +40,28 @@
...
+		[% END %]
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION" desc="" %]
+			<select name="pref_workPlaylistPosition" id="workPlaylistPosition">
+			  <option [% prefs.workPlaylistPosition == "before" ? "selected" : "" %] value="before" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_BEFORE" | string %]</option>
+			  <option [% prefs.workPlaylistPosition == "after"  ? "selected" : "" %] value="after" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER" | string %]</option>
+		          <option [% prefs.workPlaylistPosition == "hidden" ? "selected" : "" %] value="hidden" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN" | string %]</option>
+			</select>
```
michaelherger, line 58: Can you make sure you use tabs rather than spaces to indent? Thanks!

Plugin.pm
```diff
@@ -954,31 +959,74 @@ sub QobuzGetTracks {
...
 				} unless $works->{$workId};
-
+				
+				if ($workId ne $lastwork) {  # create a new work heading
+#					if ($i && $lastwork eq "") {  # create a separator line after previous single track#
+#						push @$items,{
+#							name  => "--------",
+#							type  => 'text'
+#						};
+#					}
```
michaelherger, line 969: Please remove.

Plugin.pm
```diff
@@ -954,31 +959,74 @@ sub QobuzGetTracks {
...
+					$worksfound = 1;   # we found a work with more than one track
+				}
+			} elsif ($lastwork ne "") {  # create a separator line for tracks without a work
+						push @$items,{
+							name  => "--------",
+							type  => 'text'
+						};
+						
+					$lastwork = "";
+			}
```
michaelherger, line 989: Please indent only one level.

michaelherger: And you could give an m-dash "—" a try to have a line without spaces

————————

Plugin.pm
```diff
@@ -954,31 +959,74 @@ sub QobuzGetTracks {
...
-					passthrough => [{
-						tracks => $workTracks
-					}],
-					items => $workTracks
-				} if scalar @$workTracks > 1;
+			if ( $prefs->get('workPlaylistPosition') eq "before" ) {
+				# insert work playlists before the tracks unless there is only one work containing all tracks
+				foreach my $work (sort { $works->{$b}->{index} <=> $works->{$a}->{index} } keys %$works) {
+					my $workTracks = $works->{$work}->{tracks};
+					splice @$items, 0, 0, {
```
michaelherger, line 1001: replace with `unshift @$items, {...}`

Plugin.pm
```diff
@@ -954,31 +959,74 @@ sub QobuzGetTracks {
...
+						name => $works->{$work}->{title},
+						image => 'html/images/playlists.png',
+						type => 'playlist',
+						playall => 1,
+						url => \&QobuzWorkGetTracks,
+						passthrough => [{
+							tracks => $workTracks
+						}],
+						items => $workTracks
+					} if $worksfound && scalar @$workTracks && scalar @$workTracks < $album->{tracks_count};
```
michaelherger, line 1011: A question of style... but if I have such a longish instruction and a longish condition, then I prefer to have a regular `if (cond) { block of code... }` statement.

Plugin.pm
```diff
@@ -954,31 +959,74 @@ sub QobuzGetTracks {
...
```
michaelherger, line 1028: See above comment about `if` statements.

Plugin.pm
```diff
@@ -1208,8 +1257,29 @@ sub _trackItem {
...
 	}
 
-	if ( $track->{work} ) {
-		$item->{work} = $track->{work};
+	# Enhancements to work/composer display for classical music (tags returned from Qobuz are all over the place)
+	my $genreList = $prefs->get('classicalGenres');
+	if ( $prefs->get('useClassicalEnhancements')
+		&& ( grep(/Classique/,@{$track->{album}->{genres_list}}) || grep(/(^\s*$genre\s*$|^\s*\*$)/,(split ',', $genreList )) ) 
+	   ) 
+	{
```
michaelherger, line 1265: Hmm... tricky formatting here 😁. 

I'm not sure I understand the second part of the regex. Is this to match empty entries? If that was the case then you should probably make sure you remove empty values before storing the string in the prefs. And you should probably split by a regex `/\s*,\s*` instead of just the comma to get rid of surrounding spaces.

Plugin.pm
```diff
@@ -1208,8 +1257,29 @@ sub _trackItem {
...
+		if ( $track->{work} ) {
+#			if ( $track->{work} ne $track->{title} ) {
```
michaelherger, line 1267: remove

Plugin.pm
michaelherger: And I'm not sure whether you've tested for the existence of `$track->{album}->{genre_list}` before. It might be worth a check before trying to access it.

Plugin.pm
```diff
@@ -1208,8 +1257,29 @@ sub _trackItem {
...
+				$item->{work} = $track->{work};
+#			}
+		} else {
+			# Try to extract the work from the title
+			if (index($track->{composer}->{name}, (split ':', $track->{title})[0]) == -1) {
```
michaelherger, line 1272: Again: split by regex, check for the existence of "deep" attributes like `$track->{composer}->{name}`.

ProtocolHandler.pm
```diff
@@ -269,10 +269,20 @@ sub getMetadataFor {
 
 	$meta->{title} = Plugins::Qobuz::API::Common->addVersionToTitle($meta);
 	
-	if ($meta->{work}) {
-		$meta->{title} = $meta->{work} . ': ' . $meta->{title};
+	my $genreList = $prefs->get('classicalGenres');
+	if ( $prefs->get('useClassicalEnhancements') 
+		&& ( grep(/Classique/,@{$meta->{genres_list}}) || grep(/(^\s*$meta->{genre}\s*$|^\s*\*$)/,(split ',', $genreList)) )
```
michaelherger, line 274: You should write a small helper function to do this grepping. It's pretty complicated, needs another check, and you use it more than once.

Once it's in a single place you could also consider optimizing it: pre-split the scalar genre list into a list, or even a hash for quicker checks.

strings.txt
```diff
@@ -352,3 +352,30 @@ PLUGIN_QOBUZ_PROGRESS_READ_PLAYLISTS
...
+	EN	Choose Yes here if you want the plugin to use Qobuz data to attempt to:<br />* Show a playlist and subheadings for each composer/work within a release.<br />* Add the composer and work names to the track title in the playlist and now playing displays (the playlist cannot show subheadings).
+	
+PLUGIN_QOBUZ_CLASSICAL_GENRES
+	EN	Additional classical genres
+	
+PLUGIN_QOBUZ_CLASSICAL_GENRES_DESC
+	EN	If you find a release for which the enhanced display does not show, add the Qobuz genre (shown in the release listing) to the "additional genres" list here (comma separated if more than one).
+	
+PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_TEXT
+	EN	<br /><br />For questions/bug reporting <a href="https://forums.slimdevices.com/forum/user-forums/3rd-party-software/94305-qobuz-com-streaming-plugin" target="_blank">visit the forum</a>.<br /><br /><b>Note:</b> for best results, use the "Material Skin" plugin. Otherwise, if using the default interface, do not use "large artwork" (option at bottom left of screen), the display will not be rendered well if you do.
```
michaelherger, line 369: Instead of including HTML in the string table you could use line feeds and have them replaced in the HTML template using the [`html_break`](http://www.template-toolkit.org/docs/manual/Filters.html#section_html_break_html_para_break) filter. That would give you the flexibility (if ever needed) to use the same strings in Squeezeplay or some other non-HTML interface.

michaelherger (changes requested):
Thanks a lot for this PR! I'll trust you and your testers about the logic and functionality.  Thus my review is a bit pedantic on style 😉.

Overall I think it's looking good. There are some nitpicks around formatting, and a few Perl tweaks I'd like to see. 

Oh, and please include the suggested German translation. I'll review that as well. Thanks!

Keep the good work coming!

Plugin.pm
darrell-k: I was unhappy with the whole "if (scalar keys %$works) {" block of code, it was there before, but had become a lot more complex after SamY's and then my changes, and it ended up with duplicated code because it was adding elements to the items array one at a time, which meant the splice/unshift had to be done with a reverse sort of the elements to be added. 

Your comments have pushed me to refactor it. I know I have introduced an intermediate array, but I think the code is now easier to read/maintain. What do you think?

```
		if (scalar keys %$works) {
			# create work playlists unless there is only one work containing all tracks
			my @workPlaylists = ();
			foreach my $work (sort { $works->{$a}->{index} <=> $works->{$b}->{index} } keys %$works) {
				my $workTracks = $works->{$work}->{tracks};
				if ( $worksfound && scalar @$workTracks && scalar @$workTracks < $album->{tracks_count} ) {
					push @workPlaylists, {
						name => $works->{$work}->{title},
						image => 'html/images/playlists.png',
						type => 'playlist',
						playall => 1,
						url => \&QobuzWorkGetTracks,
						passthrough => [{
							tracks => $workTracks
						}],
						items => $workTracks
					}
				}
			}
			if ( @workPlaylists ) {
				# insert work playlists according to the user preference
				if ( $prefs->get('workPlaylistPosition') eq "before" ) {
					unshift @$items, @workPlaylists;
				} elsif ( $prefs->get('workPlaylistPosition') eq "after" ) {
					push @$items, @workPlaylists;
				}
			}
		}

```

Plugin.pm
darrell-k: The second part of the regex was a hangover from before I identified that the Qobuz tag genre_list=Classique could drive the whole thing - the previous version allowed the user to add a genre of "*" in the preferences if they were overwhelmingly a classical listener and didn't want to spend the rest of their lives adding Qobuz genre tags to their list. Removed.

Using your suggested regex in the split instead of in the grep doesn't work when an additional genre entered by the user contains spaces. For example, if the user enters an additional genre of "Vocal Jazz", it will then match the Qobuz genre of "Jazz". Unlikely, but you know...

I've added a check for the existence of $track->{album}->{genre_list} to the code:
`&& ( $track->{album}->{genres_list} && grep(/Classique/,@{$track->{album}->{genres_list}}) || grep(/^\s*$genre\s*$/,(split ',', $genreList )) ) `

michaelherger: my suggestion would still require a space to be there as the separator. But it would accept " , " as well as "," or ", ".

strings.txt
darrell-k: Can you point me towards a usage of this?

michaelherger: The blue `html_break` actually is the link to the documentation 😉. 

http://www.template-toolkit.org/docs/manual/Filters.html#section_html_break_html_para_break

Plugin.pm
michaelherger: ~Two~ A few more comments on that function:

- I'd return early if `useClassicalEnhancements` was not set. Makes the `if` statement more readable: `return if !$prefs->get('useClassicalEnhancements');` as the second statement in the function. Also saves splitting genres in this case.
- `classicalGenres` could be undefined, couldn't it? In that case the `split` would complain. Make sure that pref exists before trying to split it.
- If you use a variable in a regex you should "quote" it: `/^\s*\Q$genre\E\s*$/`. Otherwise a value like `Mu$ic` would crash LMS or otherwise behave badly (https://perldoc.perl.org/perlre#Quoting-metacharacters)
- if that method was called often you could map the `@genreList` to a hash with the entries as the key and a truthy value. It then would be much more efficient looking up a genre by doing `$genreList{$genre}` instead of grepping using a regex over and over again.

Plugin.pm
michaelherger: Looks good to me. I was mostly commenting on the `splice @a, 0, 0, {}` which is technically the same as an unshift, but less readable.

Plugin.pm
darrell-k: classicalGenres is defined, but might be empty, which I know is a boolean false in Perl - is that the same as undefined? In any case, I've just tested with it an empty preference, and there are no complaints in the log.

michaelherger: Empty is not the same as undefined, no. If you initialize the pref with an empty value, you should be good.

strings.txt
darrell-k: OK, two questions on this:
1. Is it possible to apply a filter to the "desc" text string within a [% WRAPPER...]?
2. How would you deal with adding <b> and <a href...>? Separate strings?

michaelherger: Ok, you're right. My suggestion probably was a bit over-engineered. Never mind 😁.

`_DESC` elements would even be harder indeed.

darrell-k:
Hi Michael,

I've just pushed the latest version to my fork and this PR seems to have been updated accordingly.

Re previous discussion about the helper function to determine whether to use the display enhancements, I realised we only need to test for this once per album, so I moved the code to the _precacheAlbum subroutine in Common.pm.

I've also added some code to "asciify", lowercase and remove spaces from work/track strings when comparing them, in an attempt to work through Qobuz typos etc.

SamY has tested and I don't seem to have broken anything with these changes.

I think I've addressed all your comments. Let me know what you think.

API/Common.pm
```diff
@@ -107,13 +107,25 @@ sub _precacheAlbum {
 		$album->{genre} = $album->{genre}->{name};
 		$album->{image} = __PACKAGE__->getImageFromImagesHash($album->{image}) || '';
 
+		# If the user pref is for classical music enhancements to the display, is this a classical release or has the user added the genre to their custom classical list?
+		$isClassique = 0;
+		if ( $prefs->get('useClassicalEnhancements') ) {
+			my %genreList = map { $_ => 1 } split '\s*,\s*', $prefs->get('classicalGenres');
```
michaelherger, line 113: You could do this globally in the module, and update whenever there's a change to the pref. Something like:

```perl
# somewhere at the top:
my %genreList;
initGenreMap();

sub initGenreMap {
   %genreList = map { $_ => 1 } split '\s*,\s*', $prefs->get('classicalGenres');
}

$prefs->setChange(\&initGenreMap, 'classicalGenres');
```

This way you don't have to re-build the map whenever you have to pre-cache an album.

API/Common.pm
```diff
@@ -107,13 +107,25 @@ sub _precacheAlbum {
 		$album->{genre} = $album->{genre}->{name};
 		$album->{image} = __PACKAGE__->getImageFromImagesHash($album->{image}) || '';
 
+		# If the user pref is for classical music enhancements to the display, is this a classical release or has the user added the genre to their custom classical list?
+		$isClassique = 0;
+		if ( $prefs->get('useClassicalEnhancements') ) {
+			my %genreList = map { $_ => 1 } split '\s*,\s*', $prefs->get('classicalGenres');	
+			if ( $album->{genres_list} && grep(/Classique/,@{$album->{genres_list}}) || $genreList{$album->{genre}} ) {
+			$isClassique = 1;
```
michaelherger, line 115: Please indent.

API/Common.pm
```diff
@@ -107,13 +107,25 @@ sub _precacheAlbum {
...
 		my $albumInfo = {
 			title  => $album->{title},
 			id     => $album->{id},
 			artist => $album->{artist},
 			image  => $album->{image},
 			year   => (localtime($album->{released_at}))[5] + 1900,
 			goodies=> $album->{goodies},
+			genre  => $album->{genre},
+			genres_list => $album->{genres_list},
+			isClassique => $isClassique,
```
michaelherger, line 128: Wouldn't that fail to pick up changes in the `classicalGenres` list? You might need to evaluate that flag whenever you read the data.

API/Common.pm
```diff
@@ -169,6 +181,9 @@ sub precacheTrack {
 		goodies  => $album->{goodies},
 		version  => $track->{version},
 		work     => $track->{work},
+		genre    => $album->{genre},
+		genres_list => $album->{genres_list},
+		isClassique => $isClassique,
```
michaelherger, line 186: See above comment.

HTML/EN/plugins/Qobuz/settings/basic.html
```diff
@@ -40,6 +40,25 @@
 				<option [% prefs.showYearWithAlbum ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
 			</select>
 		[% END %]
+		<hr style="opacity:0.4" class="sub-sep"/>
```
michaelherger, line 43: I personally try to avoid hard-coded formatting in the templates. This should be the skin's duty to get right.

Plugin.pm
```diff
@@ -33,6 +33,8 @@ tie my %localizationTable, 'Tie::RegexpHash';
 	qr/^Livret Num.rique/i => 'PLUGIN_QOBUZ_BOOKLET'
 );
 
+$prefs->set('classicalGenres','') if (! $prefs->get('classicalGenres'));
```
michaelherger, line 36: Just add a line to the below `init()` statement:

```perl
$prefs->init({
      classicalGenres => '',
      preferredFormat...
```

Plugin.pm
```diff
@@ -1208,8 +1245,25 @@ sub _trackItem {
 		$item->{line1} .= ' (' . cstring($client, 'PLUGIN_QOBUZ_HIRES') . ')';
 	}
 
-	if ( $track->{work} ) {
-		$item->{work} = $track->{work};
+	# Enhancements to work/composer display for classical music (tags returned from Qobuz are all over the place)
+	if ( $track->{album}->{isClassique} ) {
```
michaelherger, line 1249: As outlined in a previous comment I believe you should evaluate whether an album is classical or not only here. Otherwise you might be reading a stale `isClassique` flag in case the user defined list of classical genes has changed.

Plugin.pm
```diff
@@ -1208,8 +1245,25 @@ sub _trackItem {
...
+		if ( $track->{work} ) {
+			$item->{work} = $track->{work};
+		} else {
+			# Try to extract the work from the title
+			if ( $track->{composer}->{name} && index($track->{composer}->{name}, (split ':\s*', $track->{title})[0]) == -1 ) {
```
michaelherger, line 1254: The result of the `split` statement could potentially be empty, right, thus index `0` undefined. You should evaluate this before the conditions. It also prevents doing the splitting repeatedly. 

And please put a regex in `/:\s*/` for readability.

Plugin.pm
```diff
@@ -1208,8 +1245,25 @@ sub _trackItem {
...
+				$item->{work} = (split ':\s*', $track->{title})[0];
+			} else {
+				$item->{work} = (split ':\s*', $track->{title})[1];
+			}
+                        $item->{line1} =~ s/$item->{work}://;
+		}
+		$item->{displayWork} = $item->{work};
+		$item->{displayWork} = $track->{composer}->{name} . ": " . $item->{work} if ($track->{composer}->{name});
```
michaelherger, line 1262: Another nitpick: instead of `": "` please use `string('COLON') . ' '`. Some languages (in particular French!) use a space in front of the colon.

Plugin.pm
```diff
@@ -1208,8 +1245,25 @@ sub _trackItem {
...
+		$item->{work} = $track->{composer}->{name} . ": " . $item->{work} if ($track->{composer}->{name});
+		my $composerSurname = (split ' ', $track->{composer}->{name})[-1];
```
michaelherger, line 1264: Again: don't access the result of the split unless you've verified there's a match. Otherwise you might end up with tons of warnings about undefined values in the log file.

ProtocolHandler.pm
```diff
@@ -268,11 +269,26 @@ sub getMetadataFor {
...
-	}
 
+	# user pref is for enhanced classical music display, and we have a classical release	
+	if ( $meta->{isClassique} ) {
+		# if the title doesn't already contain the work text
+		if ( $meta->{work} && index($meta->{title},$meta->{work}) == -1 ) {
+			$meta->{title} =~ s/${\(split " ", $meta->{composer})[-1]}:\s*// if $meta->{composer};
+			my $simpleWork = lc(unidecode($meta->{work}));
+			$simpleWork =~ s/\W//g;
+			my $simpleTitle = lc(unidecode($meta->{title}));
```
michaelherger, line 280: You could use `Slim::Utils::Text::ignoreCaseArticles('...', 1)`. This would also take into account the user's "ignore articles" setting.

michaelherger (approved):
I think this is good, thanks! I still have a few suggestions and some minor niggles about formatting and readability 😀. Ok?...

API/Common.pm
darrell-k: Done.

API/Common.pm
darrell-k: Done.

API/Common.pm
darrell-k: Seems to work fine as it is, both before and after implementing the initGenreMap subroutine you suggested.

API/Common.pm
darrell-k: See above.

HTML/EN/plugins/Qobuz/settings/basic.html
darrell-k: Done.

Plugin.pm
darrell-k: Done.

Plugin.pm
darrell-k: See above.

Plugin.pm
darrell-k: Done. I haven't seen an empty track title, but you never know...

Plugin.pm
darrell-k: Done.

Plugin.pm
darrell-k: Done.

ProtocolHandler.pm
darrell-k: Done.

darrell-k:
Changes now pushed to GitHub.

API/Common.pm
michaelherger: But if you set `isClassique` to false, then changed `classicalGenres` to include this track's genre after it has been cached, `isClassique` would not be updated to be truthy, would it?

ProtocolHandler.pm
```diff
@@ -268,11 +270,26 @@ sub getMetadataFor {
...
-	
-	if ($meta->{work}) {
-		$meta->{title} = $meta->{work} . ': ' . $meta->{title};
-	}
 
+	# user pref is for enhanced classical music display, and we have a classical release	
+	if ( $meta->{isClassique} ) {
+		# if the title doesn't already contain the work text
+		if ( $meta->{work} && index($meta->{title},$meta->{work}) == -1 ) {
+			$meta->{title} =~ s/${\(split " ", $meta->{composer})[-1]}:\s*// if $meta->{composer};
```
michaelherger, line 278: Hehe... I think that's one of those lines which gave Perl a bad reputation 😂. It's like four or five instructions in a single line, seasoned with some regex magic. Please move the split out of the regex and give it a name for understandability.

And whenever you use input data as part of a regex make sure you "escape" it using `/\Q$variable/`. Otherwise you risk to crash the server if eg. the value contained a slash or other meta character used in regular expressions.

https://perldoc.perl.org/functions/quotemeta

API/Common.pm
```diff
@@ -107,13 +117,24 @@ sub _precacheAlbum {
 		$album->{genre} = $album->{genre}->{name};
 		$album->{image} = __PACKAGE__->getImageFromImagesHash($album->{image}) || '';
 
+		# If the user pref is for classical music enhancements to the display, is this a classical release or has the user added the genre to their custom classical list?
+		$isClassique = 0;
+		if ( $prefs->get('useClassicalEnhancements') ) {
+			if ( $album->{genres_list} && grep(/Classique/,@{$album->{genres_list}}) || $genreList{$album->{genre}} ) {
```
michaelherger, line 123: Operator precedence can be confusing (at least it is for me). You have `&&` and `||` in the same condition. Please use braces to make sure the condition is evaluated correctly (and my poor brain understands the precedence of the expression).

Plugin.pm
```diff
@@ -1208,8 +1244,30 @@ sub _trackItem {
...
+			# Try to extract the work from the title
+			if ( $track->{composer}->{name} && $track->{title} ) {
+				my @titleSplit = split /:\s*/, $track->{title};	
+				if ( index($track->{composer}->{name}, $titleSplit[0]) == -1 ) {
+					$item->{work} = $titleSplit[0];
+				} elsif ( $titleSplit[1] ) {
+					$item->{work} = $titleSplit[1];
+				}
+			}
+			$item->{line1} =~ s/$item->{work}://;
```
michaelherger, line 1261: use `\Q` before using a variable in a regex

Plugin.pm
```diff
@@ -1208,8 +1244,30 @@ sub _trackItem {
...
+		}
+		$item->{displayWork} = $item->{work};
+		$item->{displayWork} = $track->{composer}->{name} . string('COLON') . ' ' . $item->{work} if ($track->{composer}->{name});
+		$item->{work} = $track->{composer}->{name} . string('COLON') . ' ' . $item->{work} if ($track->{composer}->{name});
+		if ( $track->{composer}->{name} ) {
+			my $composerSurname = (split ' ', $track->{composer}->{name})[-1];
+			$item->{line1} =~ s/$composerSurname://;
```
michaelherger, line 1268: use `\Q` before using a variable in a regex

michaelherger (changes requested):
Almost there... please check the regexes. Then we're ready to merge. Thanks for your patience!

API/Common.pm
darrell-k: From my reading of the code (remember Perl is a foreign language to me), it's only the raw results from Qobuz that get cached, therefore `isClassique` is re-evaluated every time - 

```
Plugin.pm
   QobuzGetTracks
      getAlbum (API.pm)
         _get (album data from cache or Qobuz)
            _precacheAlbum (Common.pm)
               evaluates $isClassique based on Qobuz/cache results & user prefs
               sets isClassique in $albumInfo map
               _precacheTracks
                  _precacheTrack  
                     sets isClassique in $meta map
```

I could be horribly wrong, of course, but it explains the behaviour I am seeing.

michaelherger: ...and then `$meta` is stored in the cache, including the `$isClassique` flag, isn't it? See Line 139 (https://github.com/pierrepoulpe/SqueezeboxQobuz/pull/32/files/3d0073e0dc6df56651b6e2f38271314103572573#diff-03c728c5508826c89b0befa221adcc73ac401f2c9dd3444bf9d27eb03586f173R193)

ProtocolHandler.pm
darrell-k: OK, I've replaced with the following, in which I've also enhanced to code to check for the entire composer name as well as only the surname. I've never seen the whole composer name in a track title, but you never know...

```
	if ( $meta->{isClassique} ) {
		# if the title doesn't already contain the work text
		if ( $meta->{work} && index($meta->{title},$meta->{work}) == -1 ) {
			# remove composer name from track title
			if ( $meta->{composer} ) {
				# full name
				$meta->{title} =~ s/\Q$meta->{composer}\E:\s*//;
				# surname only
				my $composerSurname = (split " ", $meta->{composer})[-1];
				$meta->{title} =~ s/\Q$composerSurname\E:\s*//;
			}

```
Does this look OK?

michaelherger: That does add removal of the composer's full name? But yes, looking much better - thanks!

API/Common.pm
darrell-k: OK, I think I understand what you are saying: future functionality might use the classique flag straight out of the cache and end up using a stale one. Correct?

So what do we need to do? When the user pref  genre list changes, retrieve the cache, update the flag, then write the cache back?

michaelherger: It's not future functionality - see https://github.com/pierrepoulpe/SqueezeboxQobuz/blob/master/API.pm#L511. We don't write to the cache for nothing. We use it to speed up future lookups (eg. when fetching metadata to be displayed in the UI).

I think instead of storing the flag itself you should store all the data needed to evaluate the value at run time. Then whenever you get track information, no matter whether freshly fetched from remote or out of the cache, you evaluate the flag once again.

API/Common.pm
darrell-k: Done

Plugin.pm
darrell-k: Done

Plugin.pm
darrell-k: Done

API/Common.pm
darrell-k: I must admit I'm getting a bit confused here (I often do!).  Isn't that effectively what I'm doing already? I get the Qobuz genre_list tag from their metadata (cached or not) and the additional genre list from the user Prefs, then set the isClassique flag from those data. This happens every time the browse album list or the playlist changes.

Perhaps the issue is that the isClassique flag doesn't really belong in the metadata maps, but adding it there was a very easy way of making it available in the places where it needs to be checked.

michaelherger: In ProtocolHandler you potentially get the data from the cache, incl. that flag. In that case you don't evaluate the genres any more but rely on the `isClassique` flag coming from the cache. It's probably the only place where you're not dealing with fresh data. 

We could consider this negligible. That method is called while a track is playing. If you change the genre list in your settings while an affected track is playing... oh well. 

Let's consider this good enough.

darrell-k: The behaviour I have seen during testing is that the playing item in the playlist gets updated if you change the preference to include/exclude the genre, then refresh the album listing. Then the whole playlist gets updated on any addition to the playlist, track change, or user initiated seek or skip event. This is using the Material skin, not sure if that makes a difference. 

I don't mind getting to the bottom of this (or just implementing any further changes you suggest). It's a great learning exercise. But equally, if you're happy, I'm happy, and I'll update GitHub with things as they are.

Which do you prefer?

michaelherger: I think we can leave it as it is and come back should there be an issue.

Plugin.pm
michaelherger: But not pushed yet?

darrell-k: Pushed now. Twice! ;)

michaelherger:
Release time! Thanks!

PR #33, opened 2023-04-22
Support both adding and removing artist and album favorites

SamInPgh:
Currently, the "Add to favorites" function in both the artist and album menus is unaware of the current favorite status. This change will allow toggling it by displaying either "Add to" or "Remove from" favorites, depending on the current state. Also, in order to eliminate confusion with LMS favorites, "favorites" is changed to "Qobuz favorites" in the message text.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -521,16 +521,23 @@ sub QobuzArtist {
 			passthrough => [{
 				artistId  => $artist->{id},
 			}],
-		},{
-			name => cstring($client, 'PLUGIN_QOBUZ_ADD_FAVORITE', $artist->{name}),
-			image => 'html/images/favorites.png',
-			url  => \&QobuzAddFavorite,
-			passthrough => [{
-				artist_ids => $artist->{id},
-			}],
-			nextWindow => 'parent'
 		};
-;
+		
+		Plugins::Qobuz::API->getUserFavorites(sub {
+			my $favorites = shift;
+			my $artistId = $artist->{id};
+			my $isFavorite = grep { $_->{id} eq $artistId } @{$favorites->{artists}->{items}};
+		
+			push @$items, {
+ 				name => cstring($client, $isFavorite ? 'PLUGIN_QOBUZ_REMOVE_FAVORITE' : 'PLUGIN_QOBUZ_ADD_FAVORITE', $artist->{name}),
+				url  => $isFavorite ? \&QobuzDeleteFavorite : \&QobuzAddFavorite,
+				image => 'html/images/favorites.png',
+				passthrough => [{
+					artist_ids => $artist->{id},
+				}],
+				nextWindow => 'parent'
+			}
+		});
 
 		$cb->( {
 			items => $items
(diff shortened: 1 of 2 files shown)


Plugin.pm
```diff
@@ -521,16 +521,23 @@ sub QobuzArtist {
...
+			push @$items, {
+ 				name => cstring($client, $isFavorite ? 'PLUGIN_QOBUZ_REMOVE_FAVORITE' : 'PLUGIN_QOBUZ_ADD_FAVORITE', $artist->{name}),
+				url  => $isFavorite ? \&QobuzDeleteFavorite : \&QobuzAddFavorite,
+				image => 'html/images/favorites.png',
+				passthrough => [{
+					artist_ids => $artist->{id},
+				}],
+				nextWindow => 'parent'
+			}
+		});
```
michaelherger, line 540: Unfortunately that wouldn't work (reliably): as the call might need to go online, and that lookup is asynchronous, the addition of the menu item might happen _after_ the menu has been returned to the player. You'd have to queue up these async requests and only return the menu once all the requests have been handled... welcome in callback hell!

What I've been doing in some places is create a small helper function which would return the menu (the final `$cb->()` call). That helper function would know the list of async functions. And only once all of them had responded it would actually call the `$cb->()`. 

Or, if you didn't mind making this available to LMS 8+ users only, you could use [`Async::Util::achain()`](https://metacpan.org/pod/Async::Util#achain) to simplify this. I only added this module in 8.0 (I think), in support of the Online Music Library Integration.

One more thing to keep in mind with async functions: they don't always return in the order you fired them off. Therefore `push` won't work. You'd have to somehow tell the callback at what position to insert the item (something like `splice @$items, $position, 0, $newItem` - see https://perldoc.perl.org/functions/splice).

michaelherger (changes requested):
I like the idea - thanks! But as you can see from my comment there's some more complexity to it. I think it should be reasonably simple to do correctly with LMS8. But it needs some more work. See below.

Plugin.pm
SamInPgh: Drats! I knew that was too easy. ;)  Thanks for the feedback. I'm sure I'll have more questions as I look into it further. As for requiring LMS 8+, I would have no problem with it. Could it be done with conditional code or would the plugin only be available to LMS 8+ users?

michaelherger: `if (Slim::Utils::Versions->compareVersions($::VERSION, '8.0.0') >= 0) { ... }` - see at the top of that file.

Plugin.pm
```diff
@@ -521,20 +521,27 @@ sub QobuzArtist {
...
-;
-
-		$cb->( {
-			items => $items
-		} );
+		
+		Plugins::Qobuz::API->getUserFavorites(sub {
+			my $favorites = shift;
+			my $artistId = $artist->{id};
+			my $isFavorite = grep { $_->{id} eq $artistId } @{$favorites->{artists}->{items}};
```
michaelherger, line 529: Please make sure `$favorites->{artists}->{items}` does exist and is a list before accessing it.

Plugin.pm
```diff
@@ -1020,86 +1027,91 @@ sub QobuzGetTracks {
...
-		},{
-			name => $album->{tracks_count},
-			label => 'PLUGIN_QOBUZ_TRACKS_COUNT',
-			type => 'text'
 		};
 
-		if ($album->{description}) {
+		Plugins::Qobuz::API->getUserFavorites(sub {
+			my $favorites = shift;
+			my $isFavorite = grep { $_->{id} eq $albumId } @{$favorites->{albums}->{items}};
```
michaelherger, line 1034: validate please

michaelherger (changes requested):
Nice one! With only one callback per handler it's not as bad as I thought. And no messing with LMS versions.

Please add some validation to the response from the favorites lookup.

Plugin.pm
SamInPgh: See latest commit.

Plugin.pm
SamInPgh: See latest commit.

PR #34, opened 2023-04-27
Ensure tracks in the same work are grouped even if composer metadata is missing from some.

SamInPgh:
For some reason, Qobuz does not populate the composer tag for classical tracks that are tagged as unavailable for streaming. This prevents them from being included in the 'work' object to which they belong in the plugin, which is keyed on composer and work. Instead, they are added to separate works, keyed on the work name only, which is confusing. Fortunately there are not a whole lot of releases containing these tracks, but enough to bother someone with a low threshold of annoyance such as myself.  ;-)  This change groups these tracks with the others in same work.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -949,7 +949,8 @@ sub QobuzGetTracks {
 		my $works = {};
 		my $lastwork = "";
 		my $worksfound = 0;
-		my $containsUnstreamable = 0;
+		my $noComposer = 0;
+		my $workHeadingPos = 0;
 		
 		foreach my $track (@{$album->{tracks}->{items}}) {
 			$totalDuration += $track->{duration};
@@ -960,26 +961,31 @@ sub QobuzGetTracks {
 				my $workId = Slim::Utils::Text::matchCase(Slim::Utils::Text::ignorePunct($work));
 				$workId =~ s/\s//g;
 
-				$works->{$workId} = {
+				$works->{$workId} = {   # create a new work object
 					index => $i,
-					title => $work,
-					image => $formattedTrack->{image},
+					title => $formattedTrack->{displayWork},
 					tracks => []
 				} unless $works->{$workId};
 				
 				if ($workId ne $lastwork) {  # create a new work heading
-					push @$items,{
-						name  => $work,
+					$workHeadingPos = push @$items,{
 						name  => $formattedTrack->{displayWork},
 						type  => 'text'
 					};
-						
+					
+					$noComposer = !$track->{composer}->{name};
 					$lastwork = $workId;
 				}
 				
 				push @{$works->{$workId}->{tracks}}, $formattedTrack;
 				if (scalar @{$works->{$workId}->{tracks}} > 1) {
 					$worksfound = 1;   # we found a work with more than one track
+				}
+				
+				if ($noComposer && $track->{composer}->{name} && $workHeadingPos) {  #add composer to work title if needed
+					@$items[$workHeadingPos-1]->{name} = $formattedTrack->{displayWork};
+					$works->{$workId}->{title} = $formattedTrack->{displayWork};
+					$noComposer = 0;
 				}
 			} elsif ($lastwork ne "") {  # create a separator line for tracks without a work
 				push @$items,{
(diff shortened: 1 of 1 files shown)


Plugin.pm
```diff
@@ -1274,7 +1282,6 @@ sub _trackItem {
 		}
 		$item->{displayWork} = $item->{work};
 		$item->{displayWork} = $track->{composer}->{name} . string('COLON') . ' ' . $item->{work} if ($track->{composer}->{name});
-		$item->{work} = $track->{composer}->{name} . string('COLON') . ' ' . $item->{work} if ($track->{composer}->{name});
```
michaelherger, line 1277: Is this no longer required? I don't see it set elsewhere?

SamInPgh: No, not in that format. The composer name is still prepended to `work` to create `displayWork`, which is used for display purposes. However, the tracks are now grouped by `work` only, resulting in a format that is identical to that shown in the Qobuz app. It gets tricky when the first track of a work is non-streamable though (meaning no composer). The composer name for the work must then be picked up from the first streamable track in the work, if any. I have seen some works consisting of just one non-streamable track and, for them, the `displayWork` field is the same as `work`. I'm hoping to get some feedback from Darrell on this change too but my testing looks very good thus far.

michaelherger:
@darrell-k - feedback?

SamInPgh:
Here's a screenshot from my "worst case" test album. As you probably know, the tracks with '*' are the non-streamable ones.

Screenshot_2023-04-27-01-26-18

darrell-k:
Sorry managed to reply to previous message. See above.

Plugin.pm
SamInPgh: _maybe it would be better for this change to be described as something like "ensure tracks in the same work are grouped even if composer metadata is missing from some"._

Good point. I agree.

darrell-k:
The samples won't play for me even with the option enabled - maybe something in my server settings? Do those Chopin samples play for you?

Anyway, should we have a warning line at the very top of the list "this release contains some unstreamable tracks" (whatever the user pref is set to)?

SamInPgh:
No, those samples don't play for me either and an error message is written to the server log. However, I tried a couple of other non-classical unstreamable tracks and the samples played correctly. I think it's a Qobuz problem. Because of this flaky behavior, I turned off the plugin option to play them so they won't be added to my player queue. I don't think it's worth adding a warning line though. Having the "*" before the track names should be enough for the user to figure it out. Also, there seem to be very few of these tracks, especially in the classical genre. I had a hard time finding even one other release besides that Chopin one.

darrell-k:
I amended the plug-in to show the (hopefully) relevant metadata. See screenshot. The tracks in question on the Chopin release say streamable from today(!) but still streamable=no and sampleable=no. I cleared cache.db and Qobuz.db but that made no difference.

Have you an example of an album containing playable samples?

Screenshot 2023-04-28 13 27 27

SamInPgh:
Interesting info. Unfortunately, I am having no luck finding other albums -- classical or not -- that contain non-streamable tracks. I will continue to search but they seem to be becoming rarer. I believe that Qobuz is trying to clean things up on their end. Apparently there have been many complaints from users about tracks in their saved playlists becoming unstreamable due to the albums being recatalogued with different track id's and they're attempting to address the situation.

darrell-k:
Yes, I've had that problem with a few favourites - momentary panic but then searching and re-adding to favourites solved it. Since my last post, I've discovered that the `streamable_at` date is meaningless - some favourites I added ages ago have dates this year, and even within the last month.

SamInPgh:
Okay. It took awhile but I managed to find another classical album with unstreamable tracks AND playable samples of those tracks:

Joseph Haydn: Die Schöpfung (La Création)
qobuz:album:5413633582280

The three sample tracks are 30-second mp3's.

darrell-k:
Oh, what did you think of my suggestion that we default the Enhanced Classical Display to enabled (to assist users who don't read the release notes)?

michaelherger:
Oops... my words of thanks went in to the commit message... so here again:

Thanks guys! I'll set the `useClassicalEnhancements`. As @SamInPgh mentioned this would only have an impact on users who haven't installed the plugin yet, or who have not touched the settings since the last update. I see that the value currently isn't initialized. So setting this now will enable it for many, I guess. But that's fine with me.

SamInPgh:
@darrell-k I just re-read your last post and realized that I missed the opportunity to make use of the `sampleable` tag to prevent those tracks with it set to '0' from being added to the player queue even if the plugin option to play sample tracks is selected. I'm not sure if it's worth creating another pull request but let's keep it in mind to be rolled into the next update.

PR #35, opened 2023-04-30
Allow orphaned album favorites to be removed.

SamInPgh:
This change allows album favorites whose associated album id's are not found in the Qobuz library to be removed via the Album Favorites menu, saving the user from having to perform the task in the Qobuz app.


SamInPgh:
This is the screen that will be displayed when an orphaned entry is selected from the album favorites menu:

Screenshot_2023-04-30-18-36-19

michaelherger (approved):
Nice addition!

PR #38, opened 2023-05-11
Add functionality for as-yet-unreleased albums

SamInPgh:
Allow viewing the tracklist, as well as other album information, for as-yet-unreleased albums. Also allow any streamable tracks to be played and for the album to be added or removed from user favorites.

### DIFF
--- a/API.pm
+++ b/API.pm
@@ -706,20 +706,8 @@ sub _get {
 
 	if (!$params->{_nocache} && (my $cached = $cache->get($url))) {
 		main::DEBUGLOG && $log->is_debug && $log->debug("found cached response: " . Data::Dump::dump($cached));
-		my $refresh = 0;
-		if ( $cached->{release_date_stream} ) {
-			my $today = Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d");
-			if ( $cached->{release_date_stream} ge $today ) {
-				# remove from cache in case previous version of plugin had added it, then set flag to force re-retrieve from Qobuz
-				$refresh = 1;
-				$cache->remove($url);
-			}
-		}
-
-		if ( !$refresh ) {
-			$cb->($cached);
-			return;
-		}
+		$cb->($cached);
+		return;
 	}
 
 	Slim::Networking::SimpleAsyncHTTP->new(
@@ -732,7 +720,7 @@ sub _get {
 			main::DEBUGLOG && $log->is_debug && $url !~ /getFileUrl/i && $log->debug(Data::Dump::dump($result));
 
 			if ($result && !$params->{_nocache}) {
-				if ($result->{release_date_stream} && $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) {
+				if ( !($params->{album_id}) || ( $result->{release_date_stream} && $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) ) {
 					$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
 				}
 			}
--- a/API.pm
+++ b/API.pm
@@ -705,9 +705,20 @@ sub _get {
 	}
 
 	if (!$params->{_nocache} && (my $cached = $cache->get($url))) {
-		main::DEBUGLOG && $log->is_debug && $log->debug("found cached response: " . Data::Dump::dump($cached));
-		$cb->($cached);
-		return;
+		my $refresh = 0;
+		if ( $cached->{release_date_stream} ) {
+			my $today = Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d");
+			if ( $cached->{release_date_stream} ge $today ) {
+				# remove from cache in case previous version of plugin had added it, then set flag to force re-retrieve from Qobuz
+				$refresh = 1;
+				$cache->remove($url);
+			}
+		}
+
+		if ( !$refresh ) {
+			$cb->($cached);
+			return;
+		}
 	}
 
 	Slim::Networking::SimpleAsyncHTTP->new(
@@ -720,7 +731,9 @@ sub _get {
 			main::DEBUGLOG && $log->is_debug && $url !~ /getFileUrl/i && $log->debug(Data::Dump::dump($result));
 
 			if ($result && !$params->{_nocache}) {
-				$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+				if ($result->{release_date_stream} && $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) {
+					$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+				}
 			}
 
 			$cb->($result);
--- a/API.pm
+++ b/API.pm
@@ -705,6 +705,7 @@ sub _get {
 	}
 
 	if (!$params->{_nocache} && (my $cached = $cache->get($url))) {
+		main::DEBUGLOG && $log->is_debug && $log->debug("found cached response: " . Data::Dump::dump($cached));
 		my $refresh = 0;
 		if ( $cached->{release_date_stream} ) {
 			my $today = Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d");
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1339,6 +1339,7 @@ sub _trackItem {
 			name => cstring($client, 'PLUGIN_QOBUZ_NOT_AVAILABLE'),
 			type => 'textarea'
 		}];
+		$item->{name}      = '* ' . $item->{name};
 		$item->{line1}     = '* ' . $item->{line1};
 	}
 	else {
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -969,17 +969,6 @@ sub QobuzGetTracks {
 			});
 			return;
 			
-		} elsif ($album->{released_at} > time ) {  # the album has not been released yet
-			push @$items, {
-				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED') . ' (' . Slim::Utils::DateTime::shortDateF($album->{released_at}) . ')',
-				type  => 'text'						
-			};
-			
-			$cb->({
-				items => $items,
-			}, @_ );
-			return;
-			
 		} elsif (!$album->{streamable} && !$prefs->get('playSamples')) {  # the album is not streamable
 			push @$items, {
 				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_AVAILABLE'),
@@ -990,6 +979,13 @@ sub QobuzGetTracks {
 				items => $items,
 			}, @_ );
 			return;
+		}
+
+		if ($album->{released_at} > time ) {  # the album has not been released yet
+			push @$items, {
+				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED') . ' (' . Slim::Utils::DateTime::shortDateF($album->{released_at}) . ')',
+				type  => 'text'						
+			};
 		}
 		
 		my $totalDuration = my $i = 0;
@@ -1051,7 +1047,7 @@ sub QobuzGetTracks {
 
 		}
 
-		if (scalar keys %$works) {
+		if (scalar keys %$works && ($album->{released_at} <= time) ) { # don't create work playlists for unreleased albums
 			# create work playlists unless there is only one work containing all tracks
 			my @workPlaylists = ();
 			if ( $worksfound ) {   # only proceed if a work with more than 1 contiguous track was found			
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -981,9 +981,9 @@ sub QobuzGetTracks {
 			return;
 		}
 
-		if ($album->{released_at} > time ) {  # the album has not been released yet
+		if (!_isReleased($album) ) {
 			push @$items, {
-				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED') . ' (' . Slim::Utils::DateTime::shortDateF($album->{released_at}) . ')',
+				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED') . ' (' . $album->{release_date_stream} . ')',
 				type  => 'text'						
 			};
 		}
@@ -1047,7 +1047,7 @@ sub QobuzGetTracks {
 
 		}
 
-		if (scalar keys %$works && ($album->{released_at} <= time) ) { # don't create work playlists for unreleased albums
+		if (scalar keys %$works && _isReleased($album) ) { # don't create work playlists for unreleased albums
 			# create work playlists unless there is only one work containing all tracks
 			my @workPlaylists = ();
 			if ( $worksfound ) {   # only proceed if a work with more than 1 contiguous track was found			
@@ -1126,7 +1126,7 @@ sub QobuzGetTracks {
 			}
 
 			push @$items, {
-				name  => cstring($client, 'PLUGIN_QOBUZ_RELEASED_AT') . cstring($client, 'COLON') . ' ' . Slim::Utils::DateTime::shortDateF($album->{released_at}),
+				name  => cstring($client, 'PLUGIN_QOBUZ_RELEASED_AT') . cstring($client, 'COLON') . ' ' . $album->{release_date_stream},
 				type  => 'text'
 			};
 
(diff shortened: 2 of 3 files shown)


darrell-k:
It might be worth basing the displayed release date on the `release_date_stream` tag, because `released_at` seems to be Paris time, which means that UK/US etc users will see a release date which is a day early.  For example, the Mendelssohn release we were expecting today has a released_at timestamp of 11pm today (which is midnight in France).

darrell-k:
Still not sure. Their website player shows it as available on the 12th, which might be inaccurate for people in time zones behind France, but I can see why they've done it that way - they'd get complaints, as might we!

And though it's already the 12th in most of the world, it's still not available, never mind on the 11th.

SamInPgh:
It's actually Qobuz who has it wrong, but I'm not responsible for them. The French apparently think the world revolves around Paris. The album will be available in the farthest west US time zone (Hawaii) at noon today local time. l will respond to any complaints personally with "We said it would be released today and 11:59 pm today is still today." 😉

On another note, I surprisingly received this response from Qobuz technical support regarding whether our proposed handling of as-yet-unreleased albums was allowable under our agreement with them. Short but sweet...

Screenshot_2023-05-11-15-02-42

darrell-k:
But at the other extreme, there is Australia and the far East. Paris is, time-wise, roughly in the middle. But as a European, I would say that ;-) 

Seriously though, we could presumably adjust for the user's time zone, but it's probably not worth it.

SamInPgh:
Well maybe I'm wrong but I think we ARE adjusting for the user's time zone. That's why it says it's available on May 11 in every time zone west of France and east of the international date line -- because Qobuz tagged it as available at midnight local time in France. In France and eastward to the IDL, it will say it's available on May 12.

Or am I, as often happens at my age, completely missing something?

darrell-k:
No, you're right. I just changed my server's timezone to Australia, and the release date displayed for the album showed as the 12th. Of course, that didn't make it magically available - they'll just have to wait for it to be the 12th in France. 

It will be interesting to see if it is released exactly at midnight CEST, they might just mean "sometime on the 12th" i.e. it might be a batch process that runs sometime during the night. Only an hour to go and we find out!

SamInPgh:
I'm pretty sure I ran a test awhile ago where I formatted the released_at attribute down to the second and it came out as midnight French time but, as you say, it really depends on when they actually update the library contents, which could be some time after midnight. Only 27 minutes left to wait!

darrell-k:
Yes, that's what I meant - the epoch value translates to exactly midnight CEST. But it could well be a case of spurious precision! We'll soon see...

darrell-k:
Well that was underwhelming - it's like the old days, waiting for the record shop to open.

We're now showing the full details, with the 2 sample tracks (which is what I would have seen before, with your latest changes, if I understand correctly).

SamInPgh:
Okay. It is now supposedly released but the unstreamable tracks are still unstreamable until the library is updated. I'll check every hour or so and see what's happening.

SamInPgh:
Yes, the only difference for me is there is no longer an "*" before the name and no message indicating that it is not yet released.

darrell-k:
Let me know when it becomes available. I'll check my mails in the morning,
and if no news, I'll keep checking and do the same.

darrell-k:
It's available now in the Qobuz app and web player. Still showing the preview only in the plug-in. Do we have a caching problem?

SamInPgh:
That's weird. It's still not available here in the app.

Screenshot_2023-05-11-20-13-11

SamInPgh:
The web site says it will be available to stream in about two hours.

IMG_20230511_203429

darrell-k:
Different servers over in the US, not yet updated?

Doesn't explain its non-availability in the plug-in Vs Qobuz app here. (Definitely all tracks play on my phone.) I know we cache search results, what are the rules for expiry? (I know I could look, but it's very late here - in fact, I think I'll go to bed.

SamInPgh:
They just changed it to "available in 3 hours". 🤔

It looks like they might actually stagger the release so that it becomes available at midnight in each time zone on the date corresponding to that of `released_at` in France, which is more sensible than I gave them credit for.

SamInPgh:
Now it days "available in 5 hours". 🤔

darrell-k:
This morning, the plug-in was still showing the tracks as unstreamable (except the two samples). I deleted Qobuz.db, restarted the server, and it was resolved. I'm looking at the caching code, and I think we need to change the expiry, perhaps in API/Common.pm, subroutine precacheTrack.

I restored yesterday's qobuz.db, but what I don't know how to do is update the expiry of the existing cache entry, for testing. Being able to do this would avoid having to keep finding new releases daily and only being able to test once a day.

michaelherger:
In the code which reads from the cache, you could write it back with the shorter expiry. 

But overall I'd probably only decrease the caching for those cases where we know the validity should be shorter, because the tracks' metadata (availability) might change soon.

darrell-k:
Thanks for the hint. I suppose we could check the release date when first adding an album to the cache, and set the expiry for (say) 24 hours if the release date is not yet reached. We could also consider doing something cleverer on the day of release, as we seem to have demonstrated last night that the actual hour of the release is staggered depending on the timezone of the user. Otherwise, if you check too early, you're going to have to wait another 24 hours.

darrell-k:
Or even set the expiry to the release date! It took reading back my post before the obvious dawned on me!

michaelherger:
> While we're on the subject of the cache, I noticed a little while ago that Data::Dump::dump in API.pm has some difficulty with Qobuz's boolean metadata - mostly it will return 'fix' as the value. I've confirmed that the values read from Qobuz and the cache are actually either 1 or 0.

That's not a mistake, but a feature of `Data::Dump`: if a sub-object is referenced multiple times, you'd see that `fix` in its place, and a map at the bottom where it says what it was referencing to. And as booleans are tread as objects in this case, you'd see a lot of `fix` in the dump.

SamInPgh:
So it appears that Qobuz does stagger the release across time zones such that the album is not available anywhere before the date corresponding to that of `released_at` in France or, as Darrell pointed out, the date in `release_date_stream`. I'm not sure when the update occurred on the servers over here but, as of 12:30 am this morning when I went to bed, the web site said it would be available in 7 hours (after saying 2 hours at 8 pm last night). In any event, it's available now on the app and web site.

So, aside from the caching issue, what, if anything, do we do regarding this PR? We could change the date displayed in the informational message to show the same date (`release_date_stream`) no matter where the user is located, in this case May 12. Or we could go further and change all the comparisons between `time` and `released_at` to `date` and `release_date_stream`, or we could do nothing and assume that none of the plugin users are anywhere near as concerned about this as we are (a fairly accurate assumption) and aren't refreshing their LMS client apps every few seconds, anxiously hoping to hear the opening strains of the latest Mendelssohn release (although I suppose that could happen with the upcoming Dylan album).

Anyway, I'm open to suggestions.

darrell-k:
On what date to use, I'd go with `release_date_stream`, which given the staggered release schedule we have seen, seems most accurate as well as not implying the spurious precision of an epoch value adjusted for the user's timezone.

I don't think this is over the top - I'm not thinking of users checking every few seconds, but rather making a mental or other other note of "next Thursday for that", and then it not appearing until Friday. Of course, it's still maybe not going to be there on the dot of midnight, but I think that given a release date, most people would be happy if it appeared before, say, lunchtime. 

As to the cache issue, I propose:

- If current date < release_date_stream, set cache expiry to 00:00 hours on release_date_stream, or the default, whichever is sooner.
- if current date = release_date_stream, write back to the cache with immediate expiry (or remove from cache)
- If current date > release_date_stream, set the usual 30 day expiry

What do you think?

SamInPgh:
And how about the comparisons between `time` and `released_at` that occur in multiple places to determine whether an album or track should be flagged as unavailable. For consistency, should we also change those comparisons to date only? I guess we'd have to in order to keep everything in sync display-wise.

darrell-k:
Yes, I think so - grepping `released_at` indicates it wouldn't be a massive job.

SamInPgh:
Right. About five instances, as I recall. Okay, I'll take care of that part. Your idea on the handling the cache expiry sounds good. Since you're already in there, would you mind making that change on a separate PR? Either that or just send me your changes to incorporate into this one. Btw, my cached info still hasn't been updated for the Mendelssohn release.

darrell-k:
You'll probably need to wait 30 days from when you first browsed it. Or just delete qobuz.db!

I'll have a go at the cache thing (a new area for me), but it would make sense to include it in this PR.

SamInPgh:
I think we're all set now except for the caching changes. Fyi, the album release date is now displayed in YYYY-MM-DD format, straight from `release_date_stream`, rather than 'Month DD, YYYY'. If there is any objection, I'll reformat it.

darrell-k:
As I often do, I've been having second thoughts on this. It's not going to handle the case when a release date is brought forward after the user has initially browsed the release. And updating the cache expiry to the release date every time the user browses something not yet released is redundant.

How about simply not caching it at all before the release date? That would handle a brought-forward release date as well as massively simplifying the code, I think.

SamInPgh:
> OK, attached is a new API.pm (renamed API.txt) [API.txt](https://github.com/pierrepoulpe/SqueezeboxQobuz/files/11478951/API.txt) with the caching changes. I've left in some extra info messaging for testing and a (commented out) bit which I used to manipulate the release date, in case you want to run some tests of your own.
> 
> Any ideas for improving this code would be welcome - it was my first encounter with Perl date handling, which I found a bit mad - zero-based month numbers!
> 
> One thing that is missing is that it won't immediately help users who already have originally unreleased but now released albums in their cache. It will sort itself out within 30 days, of course, but is there any facility to run a routine during installation which would trawl the user's cache and remove any albums released within 30 days of installation?

Regarding the coding changes involving date handling, my understanding is that the `ymd` method is only applicable to objects of type `DateTime`. You could either change the type of $t from ` localtime` to `DateTime`  or change the `ymd` method of obtaining the current date to use:
`Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d")`

Being fairly new to Perl myself, I might be completely wrong about the above. On the other hand, it might be moot because I really really like the idea presented in your 'second thought' post (being no stranger to second guessing my own work, mostly as I'm laying in bed right after finishing a late-night coding session). It would drastically reduce the resource overhead and just makes sense on every level that I can think of. Giving it a little thought, I would suggest taking a conservative approach by waiting until the day AFTER the release date to cache the album, considering what we saw with the recent new releases where the library wasn't updated on the servers in the US until several hours after midnight here in the eastern time zone. As you said, this approach would also drastically simplify the coding as well as the overhead. I'm all for it!

michaelherger:
May I ask how often you'd encounter those unreleased works?

darrell-k:
Sam:

The date stuff works (according to my testing) - I used https://perldoc.perl.org/Time::Piece .

Michael:

I don't know the answer to your question. But the ratio of released/unreleased albums must be very large (and growing) so statistically you would encounter an unreleased album very infrequently. I suppose we could cut the cache expiry to 1 day, rather than remove it completely, if you are worried about this change's effect on the agreement with Qobuz.

michaelherger:
I'm not really worried. I was only wondering whether the effort you put into this was worth it. 

As for `Time::Piece`: try to avoid adding modules, in particular binaries which you'd have to provide for each and every Perl version/platform/operating system permutation the users might be using. You should be able to do what you want to do with those modules already there.

darrell-k:
Well, the effort is already put in now. I think it was worth it - the use case of searching for upcoming releases is not an obscure one, IMO.

I'll go ahead with the simplification, that way we'll not need the Time::Piece module, but I'll bear in mind your advice on new modules for the future.

darrell-k:
Here is a new version of API.pm (renamed to API.txt) which simply does not cache albums until the day after release date. It also resolves the problem where a user has an unreleased album in their cache due to a previous version of the plug-in - the album will be removed from their cache and re-retrieved from Qobuz when they access it, so that they immediately see any changes.

I have removed my debug code and also attached my testing results.

Sam, please could you add this change to the PR, so we can see that it all hangs together.

Darrell

[API.txt](https://github.com/pierrepoulpe/SqueezeboxQobuz/files/11489054/API.txt)
[release_date_cache_testing.txt](https://github.com/pierrepoulpe/SqueezeboxQobuz/files/11489058/release_date_cache_testing.txt)

darrell-k:
Ahhh! old line 708 should not have been removed - obviously I deleted it by mistake while removing my own debug code. Please could you put it back?

Thanks,
Darrell

SamInPgh:
I was wondering about that. Will do.

SamInPgh:
Committed and running with all the changes now.

SamInPgh:
I think I found a problem. I noticed that performance seemed to be suffering when running with the caching change and it appears to me that caching is now only being done for album items that have release dates in the past, as intended. However, it also appears that those items and ONLY those items are being cached. None of the other types of of items, i.e. artists, playlists, tracks, etc., are being cached at all. I have changed line 735 of my copy of Api.pm as follows:

Old:
`				if ($result->{release_date_stream} && $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) {
`

New:
`				if ( !($result->{release_date_stream}) || $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) {`

Beyond that, I'm a little uncomfortable with using the existence of `release_date_stream` to determine if the current record is an album record but, if that is the intent and if it works, it's no less efficient than checking the url type or a flag passed in the arguments. It's just not real clear to the casual observer what is being done there. I have to say, though, that it's 1:30 am here right now so nothing is real clear to me right now. Anyway, let me know what you think and don't hesitate to tell me if I'm completely missing something because... did I mention it's 1:30 am?  ;-)

SamInPgh:
Looks good, Darrell. We'll be able to get some real-world testing on all of the changes in the PR late tomorrow and early Friday. In the meantime, I'll try to find some more titles that are scheduled for release this week.

API.pm
```diff
@@ -706,8 +706,20 @@ sub _get {
...
+				# remove from cache in case previous version of plugin had added it, then set flag to force re-retrieve from Qobuz
+				$refresh = 1;
+				$cache->remove($url);
+			}
+		}
+
+		if ( !$refresh ) {
+			$cb->($cached);
+			return;
+		}
```
michaelherger, line 722: I'm not sure whether this is read for review already? But instead of adding specific code for all kinds of backwards compatibility issues you could simply bump the cache version number in API/Common.pm (https://github.com/pierrepoulpe/SqueezeboxQobuz/blob/fcc34d6e572af928cf16f61df034d64d0fcadf8c/API/Common.pm#L51): by increasing that number 2 to 3 you'd basically invalidate all of the cache.

I just don't like the idea of starting to add cache cleanup code for all kinds of incompatible changes. The cache should be considered something temporary, which can go away at any time.

darrell-k: I can see that this is much cleaner. Might lead to a bit of a spike in Qobuz server access on release day.

Not quite ready for full review/merge yet, Sam has the new version with extra debug code for testing and will commit when he is happy. 

(I have some enhancements to classical playlist handling in my fork, I suppose I should learn how to manage multiple branches rather than attaching code to messages.)

SamInPgh:
Ready for review.

Plugin.pm
```diff
@@ -1130,7 +1126,7 @@ sub QobuzGetTracks {
 			}
 
 			push @$items, {
-				name  => cstring($client, 'PLUGIN_QOBUZ_RELEASED_AT') . cstring($client, 'COLON') . ' ' . Slim::Utils::DateTime::shortDateF($album->{released_at}),
+				name  => cstring($client, 'PLUGIN_QOBUZ_RELEASED_AT') . cstring($client, 'COLON') . ' ' . $album->{release_date_stream},
```
michaelherger, line 1129: Is the `release_date_stream` string localized? Or unpredictable? Otherwise it would be nice if it could be localized for the user. I'm always confused by "2023-06-05"... is this in May or June?...

michaelherger:
Thanks guys!

Plugin.pm
SamInPgh: I believe that it's always in YYYY-MM-DD format. It could easily be localized with a few additional instructions if that would be preferable.

michaelherger: Yes, please. I can't get used to any other date format than "ours" 😀.

SamInPgh:
Please hold off on creating a new release until my next PR, which will be coming as soon as Darrell's PR40 has been merged (I don't fully trust Github yet). ;)  It will contain the change to localize the release dates and some additional enhancements that should make the next release one that I feel very confident and happy about. Hopefully, it will be the last one for awhile. I think we're all ready for a break.

Plugin.pm
SamInPgh: Coming in the next PR.

PR #39, opened 2023-05-19
Handle genre correctly when not browsing by album

darrell-k:
When not adding tracks to the playlist via the album, $meta->genre ends up being a map (with strange elements like "color"!) rather than a simple string. This fixes that.


michaelherger:
Thanks! I'm sorry for asking for a separate PR. This was almost not worth it 😁.

PR #40, opened 2023-05-21
Classical Playlist Enhancements

darrell-k:
The user can now choose to integrate playlists with work headings. So that instead of a separate block of playlists above or below the track listing, the work subheading becomes the playlist.

There are also various improvements to the slicing, recombining and handling of tags, as a result of our never-ending discovery of new types of edge case. The main one here is making sure, when we find multiple works with the same name but different composer in a release, that they generate separate subheadings in the track listing and are shown correctly in the play queue.

### DIFF
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -51,7 +51,7 @@
 			<select name="pref_workPlaylistPosition" id="workPlaylistPosition">
 				<option [% prefs.workPlaylistPosition == "before" ? "selected" : "" %] value="before" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_BEFORE" | string %]</option>
 				<option [% prefs.workPlaylistPosition == "after"  ? "selected" : "" %] value="after" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER" | string %]</option>
-				<option [% prefs.workPlaylistPosition == "hidden" ? "selected" : "" %] value="hidden" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN" | string %]</option>
+				<option [% prefs.workPlaylistPosition == "integrated" ? "selected" : "" %] value="integrated" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_INTEGRATED" | string %]</option>
 			</select>
 		[% END %]
 		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_CLASSICAL_GENRES" desc="PLUGIN_QOBUZ_CLASSICAL_GENRES_DESC" %]
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -40,6 +40,25 @@
 				<option [% prefs.showYearWithAlbum ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
 			</select>
 		[% END %]
+		<hr class="sub-sep"/>
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS" desc="PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_DESC" %]
+			<select name="pref_useClassicalEnhancements" id="useClassicalEnhancements">
+				<option [% prefs.useClassicalEnhancements ? "selected" : "" %] value="1" >[% "YES" | string %]</option>
+				<option [% prefs.useClassicalEnhancements ? "" : "selected" %] value="0" >[% "NO" | string %]</option>
+			</select>
+		[% END %]
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION" desc="" %]
+			<select name="pref_workPlaylistPosition" id="workPlaylistPosition">
+				<option [% prefs.workPlaylistPosition == "before" ? "selected" : "" %] value="before" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_BEFORE" | string %]</option>
+				<option [% prefs.workPlaylistPosition == "after"  ? "selected" : "" %] value="after" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER" | string %]</option>
+				<option [% prefs.workPlaylistPosition == "hidden" ? "selected" : "" %] value="hidden" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN" | string %]</option>
+			</select>
+		[% END %]
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_CLASSICAL_GENRES" desc="PLUGIN_QOBUZ_CLASSICAL_GENRES_DESC" %]
+			<input type="text" class="stdedit" name="pref_classicalGenres" id="classicalGenres" value="[% prefs.classicalGenres %]" size="80">
+		[% "PLUGIN_QOBUZ_USE_CLASSICAL_ENHANCEMENTS_TEXT" | string %]
+		[% END %]
+		<hr class="sub-sep"/>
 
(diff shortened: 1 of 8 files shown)


Plugin.pm
```diff
@@ -1005,32 +1012,74 @@ sub QobuzGetTracks {
...
-					tracks => []
-				} unless $works->{$workId};
-
-				if ($workId ne $lastwork) {  # create a new work heading
+				# Extended Work ID: will usually not change, but we need to keep non-contiguous tracks from the same work
+				# separate if the user has chosen to integrate playlists with the work titles.				
+				$worksWorkIdE = $worksWorkId;					
+				if ( $workPlaylistPos eq "integrated" && $works->{$worksWorkId} ) {
+					if ( $worksWorkId ne $lastWorksWorkId ) {
+						$discontigWorks->{$worksWorkId} = $worksWorkId . $i;
```
michaelherger, line 1034: Is it time to name `$i` something meaningful?...

Plugin.pm
```diff
@@ -995,6 +995,13 @@ sub QobuzGetTracks {
...
 		my $workPlaylistPos = $prefs->get('workPlaylistPosition');
+		my $currentComposer = "";
+		my $lastComposer = "";
+		my $worksWorkId = "";
+		my $worksWorkIdE = "";
+		my $lastWorksWorkId = "";
+		my $discontigWorks;
+		my $workComposer;
 
 		foreach my $track (@{$album->{tracks}->{items}}) {
```
michaelherger, line 1006: IIRC the works handling is optional, but it seems this code is always executed? Can the processing be made optional, too? I could imagine that this has become pretty heavy on CPU if iterated many times. Maybe you could even move all of the processing into it's own `sub`, to refactor things into manageable and more readable chunks of code?

michaelherger (commented):
Would be nice if you could move all (or most of it anyway) of the works handling into its own `sub`, and clearly make it optional. Remember that you can return an array of results in case you needed multiple values in the main routine:

```perl
my ($a, $b, $c) = myFunc();
```

HTML/EN/plugins/Qobuz/settings/basic.html
```diff
@@ -51,7 +51,7 @@
 			<select name="pref_workPlaylistPosition" id="workPlaylistPosition">
 				<option [% prefs.workPlaylistPosition == "before" ? "selected" : "" %] value="before" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_BEFORE" | string %]</option>
 				<option [% prefs.workPlaylistPosition == "after"  ? "selected" : "" %] value="after" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER" | string %]</option>
-				<option [% prefs.workPlaylistPosition == "hidden" ? "selected" : "" %] value="hidden" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN" | string %]</option>
+				<option [% prefs.workPlaylistPosition == "integrated" ? "selected" : "" %] value="integrated" >[% "PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_INTEGRATED" | string %]</option>
```
michaelherger, line 54: This would be a breaking change to previous revisions. Is there any migration possible?

strings.txt
```diff
@@ -393,10 +393,10 @@ PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_AFTER
...
 	NL	Na afspeellijst
 
-PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_HIDDEN
-	EN	Do not show
-	DE	Nicht Anzeigen
-	NL	Niet tonen
+PLUGIN_QOBUZ_WORK_PLAYLIST_POSITION_INTEGRATED
+	EN	Integrate with work headings
+	DE	Mit Werktiteln integrieren
+	NL	integreren met werktitels
```
michaelherger, line 399: Should probably be upper case?

Plugin.pm
darrell-k: OK, I've changed it to `$trackNumber` and then `$workNumber` (it was used in two different loops) in the `QobuzGetTracks` subroutine.

There is also a `$i` in subroutine `_getGoodiesCLI` and I'm not certain of its meaning there - goodieNumber?

michaelherger: I would often use `$i` in simple loops, where its meaning is relatively obvious. But in the above case the loop has grown considerably...

Plugin.pm
darrell-k: The works handling was there before we started on the classical enhancements road, so it wasn't optional. It's just that it only kicked in when Qobuz had populated work, which we've only ever seen for classical releases.

 The Classical Enhancements project was mainly about deriving sensible works tags when Qobuz didn't provide them. That code is restricted to classical genres only and controlled by the user pref, but the works processing, as such, never was. 

Having said that, while the pre-Classical Enhancements version of the plug-in created work heading playlists when Qobuz provided a work tag, we no longer seem to do so if the preference is switched off, so I'm investigating further.

As to a separate subroutine, I'm not sure I agree - I think the code is easier to read inline, and it's quite well demarcated in the "if works..." block. But maybe we should head that up with a comment explaining what the block is doing.

HTML/EN/plugins/Qobuz/settings/basic.html
darrell-k: I did test this: if the user has the "hidden" option selected, it just behaves the same as it did before, until they go to preferences and re-save. 

But perhaps it would be better if we automatically updated "hidden" to "integrated"?

strings.txt
darrell-k: I'll do this - for all languages?

michaelherger: I was only referring to the Dutch `integreren`, sorry.

HTML/EN/plugins/Qobuz/settings/basic.html
michaelherger: If it behaves as before, then that's fine.

Plugin.pm
michaelherger: As you like. Was only a suggestion. I personally don't like long chunks of code. Even if a section is only used once, by putting it inside its own function I feel it's cleaner, or at least should be easier to keep things in one place (eg. accidentally overwriting/re-using variables).

ProtocolHandler.pm
```diff
@@ -293,8 +294,13 @@ sub getMetadataFor {
 			}
 		}
 		
-		if ( $meta->{composer} && index($meta->{title},(split " ", $meta->{composer})[-1]) == -1 ) {
-			$meta->{title} =  (split " ", $meta->{composer})[-1] . string('COLON') . ' ' . $meta->{title};
+		# Prepend composer surname to title unless it's at the beginning the work/title (code above only strips out composer+COLON
+		# and we've found tracks with the composer name in the body of the title - we should still prepend composer to these.
+		if ( $meta->{composer} ) {
+			$composerSurname = (split " ", $meta->{composer})[-1];
```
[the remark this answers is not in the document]
michaelherger: It's required in both places, but defined in both places, too. IMHO it would be cleaner to keep the variable local. There's no reason to share it.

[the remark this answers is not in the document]
darrell-k: Sorry, I am not following this. To my mind it is defined only once (line 276) and then populated at either 284 or 300, to the same thing, `(split " ", $meta->{composer})[-1]`. It is local to the `$meta->{isClassique}` block. Sorry if I'm missing the point here.

michaelherger:
Do you have more changes to come?

darrell-k:
I'll be updating the PR, hopefully later today.

darrell-k:
Hi Michael,

In an earlier PR you suggested I replace the separator line `"--------"` with an m-dash "—", which I did, by pasting the m-dash character from your comment into the code I must admit I didn't test it, and I don't think in use we ever execute that bit of code because we've always managed to derive a work at that point. 

But in my testing today, I managed to trigger that code, and in the web interface got `ââââââââ `displayed instead. Any ideas?

michaelherger:
Ouch... the web UI still isn't fully UTF8. That might become complicated... You could potentially replace it with `&mdash;`s when `$params->{isWeb}` is set. I'm not 100% this is available where you do this.

But if this hasn't been seen before, because it's a rare case, then it's probably not worth the tweak anyway.

darrell-k:
I think I'll revert to the standard dash.

michaelherger:
It's in! Thanks!

PR #41, opened 2023-05-22
Improve handling and display of album release dates

SamInPgh:
These changes simplify how the release date and year are calculated and displayed, removing discrepancies  between the YEAR tag and the release date in time zones east of France for albums released on January 1, among other things. It also now displays album release dates using the local date format. Note: It would be nice to have a version of Slim::Utils:DateTime::shortDateF() that took a localtime structure instead of an epoch time as an argument.

The change to strings.txt corrects the statement that sample tracks are 60 seconds. They are now 30 seconds.

Albums that are not yet released but streamable due to having one or more pre-released playable tracks are no longer filtered out in Common.pm when the user has chosen not to play 30-second sample tracks, allowing those albums to be shown and the pre-release tracks to be played.

### DIFF
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -91,9 +91,8 @@ sub filterPlayables {
 
 	return $items if $prefs->get('playSamples');
 
-	my $t = time;
 	return [ grep {
-		($_->{released_at} ? $_->{released_at} <= $t : 1) && $_->{streamable};
+		!$_->{released_at} || $_->{streamable};  # allow all tracks and streamable albums
 	} @$items ];
 }
 
@@ -130,7 +129,7 @@ sub _precacheAlbum {
 			id     => $album->{id},
 			artist => $album->{artist},
 			image  => $album->{image},
-			year   => (localtime($album->{released_at}))[5] + 1900,
+			year   => substr($album->{release_date_stream},0,4),
 			goodies=> $album->{goodies},
 			genre  => $album->{genre},
 			genres_list => $album->{genres_list},
@@ -186,7 +185,7 @@ sub precacheTrack {
 		performers => $track->{performers} || '',
 		cover    => $album->{image},
 		duration => $track->{duration} || 0,
-		year     => $album->{year} || (localtime($album->{released_at}))[5] + 1900 || 0,
+		year     => $album->{year} || substr($album->{release_date_stream},0,4) || 0,
 		goodies  => $album->{goodies},
 		version  => $track->{version},
 		work     => $track->{work},
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -5,6 +5,7 @@ use base qw(Slim::Plugin::OPMLBased);
 
 use JSON::XS::VersionOneAndTwo;
 use Tie::RegexpHash;
+use POSIX qw(strftime);
 
 use Slim::Formats::RemoteMetadata;
 use Slim::Player::ProtocolHandlers;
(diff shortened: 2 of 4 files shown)


API/Common.pm
```diff
@@ -91,9 +91,8 @@ sub filterPlayables {
 
 	return $items if $prefs->get('playSamples');
 
-	my $t = time;
 	return [ grep {
-		($_->{released_at} ? $_->{released_at} <= $t : 1) && $_->{streamable};
```
michaelherger, line 96: This would let slip tracks which a release date in the future. Does that not exist?

Plugin.pm
```diff
@@ -982,8 +983,10 @@ sub QobuzGetTracks {
 		}
 
 		if (!_isReleased($album) ) {
+			my @dt = split(/-/, $album->{release_date_stream});
+			my $rDate = strftime(preferences('server')->get('shortdateFormat'), 0, 0, 0, $dt[2], $dt[1] - 1, $dt[0] - 1900);
```
michaelherger, line 987: Please create a small helper function for this. Otherwise a fix will always need to be applied to two places.

Plugin.pm
```diff
@@ -1210,7 +1215,7 @@ sub _albumItem {
 	my $artist = $album->{artist}->{name} || '';
 	my $albumName = $album->{title} || '';
 	my $showYearWithAlbum = $prefs->get('showYearWithAlbum');
-	my $albumYear = $showYearWithAlbum ? $album->{year} || (localtime($album->{released_at}))[5] + 1900 || 0 : 0;
+	my $albumYear = $showYearWithAlbum ? $album->{year} || substr($album->{release_date_stream},0,4) || 0 : 0;
```
michaelherger, line 1218: Might be worth a `getYearFromReleaseDate()` function - you're doing the same conversion several times.

michaelherger (changes requested):

API/Common.pm
SamInPgh: No. They don't exist. This function is actually called for both` track` and `album` objects. However, as I discovered previously, the `released_at `attribute only exists for albums. The effect of the old code was to filter out unstreamable tracks and albums, as well as albums that were not yet released, unless the user had selected the prefs option to play sample tracks (`playSamples`) in which case nothing was filtered out (not sure how much sense that made actually).  This change results in only unstreamable albums being filtered out at this level, allowing all users to see the contents of as-yet-unreleased albums containing streamable tracks and play those tracks. Sample tracks, of course, will only be playable if the `playSamples` option was chosen. I guess a simpler way of saying this is that users who simply don't want to play 30-second mp3 sample tracks will no longer be prevented from seeing upcoming album releases that contain streamable pre-release tracks. (Well that was at least a shorter way, if not simpler).  😊

Plugin.pm
SamInPgh: I respectfully disagree, as these are all one-line changes and the usage is really a simple assignment now rather than a conversion. The overhead for adding a function call wouldn't be worth it.

Plugin.pm
SamInPgh: Done.

Plugin.pm
michaelherger: I count six occurrences. Had I (or you) made this a helper function you'd only have had to change it once. Next time you'll have to do the same thing six times again. Unless you've already missed an occurrence 😉. But let's not get stuck on such details.

PR #42, opened 2023-05-27
Prevent import scan from caching unreleased albums

SamInPgh:
The import scan for OMLI needs to bypass the caching of as-yet-unreleased favorite albums, as is now done elsewhere. An oversight in PR #38.

### DIFF
--- a/API/Sync.pm
+++ b/API/Sync.pm
@@ -308,7 +308,9 @@ sub _get {
 		main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump($result));
 
 		if ($result && !$params->{_nocache}) {
-			$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+			if ( !($params->{album_id}) || ( $result->{release_date_stream} && $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) ) {
+				$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+			}
 		}
 
 		return $result;


API/Sync.pm
```diff
@@ -308,7 +308,9 @@ sub _get {
 		main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump($result));
 
 		if ($result && !$params->{_nocache}) {
-			$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+			if ( !($params->{album_id}) || ( $result->{release_date_stream} && $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) ) {
+				$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+			}
```
michaelherger, line 313: Is it time to factor this out into its own little helper in API::Common? It's identical in the two places, and as you've seen it's too easy to fix one, but forget the other.

SamInPgh: Sounds reasonable. Will do.

SamInPgh: I hate to waffle on this but, after looking at the code and considering how much of the `_get()` function (including the name itself) is already duplicated between the versions in API.pm and Sync.pm, and that the Sync.pm `_get()` code is exercised so little in comparison with the API.pm version, I again feel that the overhead incurred by a function call would outweigh any maintenance savings due to having one place rather than two to change it in the future, assuming that it will ever again need to be changed. 

(Btw, I'm somewhat famous in my family for the length of my sentences but cannot compete with my hero in that regard, the late author David Foster Wallace, who penned many that were multiple pages long.)   ;-)

michaelherger: What overhead? Writing that phrase must have taken longer than factoring out a little function 😂. I don't really understand the problem with the suggested change, in particular as we've seen how easily doing the same change in two places can be forgotten...

PR #43, opened 2023-05-28
Allow Qobuz track replaygain values to be used by LMS.

SamInPgh:
The plugin will now support replay gain for both tracks and albums, according to the prefs option chosen for the player. Logic to support `Smart gain` does not exist at the plugin level and this option will be treated as `Album gain` for Qobuz content.

### DIFF
--- a/API.pm
+++ b/API.pm
@@ -502,7 +502,7 @@ sub getTags {
 sub getTrackInfo {
 	my ($class, $cb, $trackId) = @_;
 
-	$cb->() unless $trackId;
+	return $cb->() unless $trackId;
 
 	if ($trackId =~ /^http/i) {
 		$trackId = $cache->get("trackId_$trackId");
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -135,6 +135,7 @@ sub _precacheAlbum {
 			genres_list => $album->{genres_list},
 			isClassique => $isClassique,
 			duration => 0,
+			isAlbumTrack => 1,
 		};
 
 		_precacheTracks([ map {
@@ -199,6 +200,7 @@ sub precacheTrack {
 		genre    => $album->{genre},
 		genres_list => $album->{genres_list},
 		isClassique => $isClassique,
+		isAlbumTrack => $album->{isAlbumTrack},
 	};
 
 	if ($track->{audio_info}) {
@@ -218,7 +220,7 @@ sub precacheTrack {
 			}
 		}
 	}
-	
+
 	$album->{duration} += $meta->{duration};
 	main::DEBUGLOG && $log->is_debug && $log->debug("Track $meta->{title} precached");
 	$cache->set('trackInfo_' . $track->{id}, $meta, ($meta->{duration} ? QOBUZ_DEFAULT_EXPIRY : QOBUZ_EDITORIAL_EXPIRY));
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -140,6 +140,7 @@ sub _precacheAlbum {
 			$_->{album} = $albumInfo;
 			$_;
 		} @{$album->{tracks}->{items}} ]);
+
 	}
 
 	return $albums;
@@ -173,7 +174,7 @@ sub precacheTrack {
 
 	my $album = $track->{album} || {};
 	$track->{composer} ||= $album->{composer} || {};
-	
+
 	my $meta = {
 		title    => $track->{title} || $track->{id},
 		album    => $album->{title} || '',
@@ -195,11 +196,20 @@ sub precacheTrack {
 	};
 
 	if ($track->{audio_info}) {
+		my $updateAlbumGain = 0;
 		if (defined $track->{audio_info}->{replaygain_track_gain}) {
 			$meta->{replay_gain} = $track->{audio_info}->{replaygain_track_gain};
+			if (!defined $album->{replay_gain} || ($album->{replay_gain} > $meta->{replay_gain})) {
+				$updateAlbumGain = 1;
+				$album->{replay_gain} = $meta->{replay_gain};
+			}
 		}
+
 		if (defined $track->{audio_info}->{replaygain_track_peak}) {
 			$meta->{replay_peak} = $track->{audio_info}->{replaygain_track_peak};
+			if ($updateAlbumGain) {
+				$album->{replay_peak} = $meta->{replay_peak};
+			}
 		}
 	}
 
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -135,7 +135,6 @@ sub _precacheAlbum {
 			genres_list => $album->{genres_list},
 			isClassique => $isClassique,
 			duration => 0,
-			isAlbumTrack => 1,
 		};
 
 		_precacheTracks([ map {
@@ -200,7 +199,6 @@ sub precacheTrack {
 		genre    => $album->{genre},
 		genres_list => $album->{genres_list},
 		isClassique => $isClassique,
-		isAlbumTrack => $album->{isAlbumTrack},
 	};
 
 	if ($track->{audio_info}) {
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -141,6 +141,11 @@ sub _precacheAlbum {
 			$_;
 		} @{$album->{tracks}->{items}} ]);
 
+		if (defined $albumInfo->{replay_gain}) {
+			$album->{replay_gain} = $albumInfo->{replay_gain};
+			$album->{replay_peak} = $albumInfo->{replay_peak};
+			$cache->set('albumInfo_' . $albumInfo->{id}, $albumInfo, QOBUZ_DEFAULT_EXPIRY);
+		}
 	}
 
 	return $albums;
@@ -212,7 +217,7 @@ sub precacheTrack {
 			}
 		}
 	}
-
+	main::DEBUGLOG && $log->is_debug && $log->debug("Track $meta->{title} precached");
 	$cache->set('trackInfo_' . $track->{id}, $meta, ($meta->{duration} ? QOBUZ_DEFAULT_EXPIRY : QOBUZ_EDITORIAL_EXPIRY));
 
 	return $meta;
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -134,6 +134,7 @@ sub _precacheAlbum {
 			genre  => $album->{genre},
 			genres_list => $album->{genres_list},
 			isClassique => $isClassique,
+			duration => 0,
 		};
 
 		_precacheTracks([ map {
@@ -217,6 +218,8 @@ sub precacheTrack {
 			}
 		}
 	}
+	
+	$album->{duration} += $meta->{duration};
 	main::DEBUGLOG && $log->is_debug && $log->debug("Track $meta->{title} precached");
 	$cache->set('trackInfo_' . $track->{id}, $meta, ($meta->{duration} ? QOBUZ_DEFAULT_EXPIRY : QOBUZ_EDITORIAL_EXPIRY));
 
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -194,8 +194,13 @@ sub precacheTrack {
 		isClassique => $isClassique,
 	};
 
-	if ($track->{audio_info} && defined $track->{audio_info}->{replaygain_track_gain}) {
-		$meta->{replay_gain} = $track->{audio_info}->{replaygain_track_gain};
+	if ($track->{audio_info}) {
+		if (defined $track->{audio_info}->{replaygain_track_gain}) {
+			$meta->{replay_gain} = $track->{audio_info}->{replaygain_track_gain};
+		}
+		if (defined $track->{audio_info}->{replaygain_track_peak}) {
+			$meta->{replay_peak} = $track->{audio_info}->{replaygain_track_peak};
+		}
 	}
 
 	$cache->set('trackInfo_' . $track->{id}, $meta, ($meta->{duration} ? QOBUZ_DEFAULT_EXPIRY : QOBUZ_EDITORIAL_EXPIRY));
--- a/API/Sync.pm
+++ b/API/Sync.pm
@@ -308,7 +308,9 @@ sub _get {
 		main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump($result));
 
 		if ($result && !$params->{_nocache}) {
-			$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+			if ( !($params->{album_id}) || ( $result->{release_date_stream} && $result->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) ) {
+				$cache->set($url, $result, $params->{_ttl} || QOBUZ_DEFAULT_EXPIRY);
+			}
 		}
 
 		return $result;
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -85,6 +85,11 @@
 			<input type="checkbox" [% IF prefs.pref_playSamples %]checked [% END %] class="stdedit" name="pref_playSamples" id="pref_playSamples" value="1" />
 			<label for="pref_playSamples" class="stdlabel">[% "PLUGIN_QOBUZ_PLAY_SAMPLES_DESC" | string %]</label>
 		[% END %]
+
+		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_USE_REPLAY_GAIN" desc="" %]
+			<input type="checkbox" [% IF prefs.pref_useReplayGain %]checked [% END %] class="stdedit" name="pref_useReplayGain" id="pref_useReplayGain" value="1" />
+			<label for="pref_useReplayGain" class="stdlabel">[% "PLUGIN_QOBUZ_USE_REPLAY_GAIN_DESC" | string %]</label>
+		[% END %]
 	[% END %]
 
 [% PROCESS settings/footer.html %]
--- a/HTML/EN/plugins/Qobuz/settings/basic.html
+++ b/HTML/EN/plugins/Qobuz/settings/basic.html
@@ -85,11 +85,6 @@
 			<input type="checkbox" [% IF prefs.pref_playSamples %]checked [% END %] class="stdedit" name="pref_playSamples" id="pref_playSamples" value="1" />
 			<label for="pref_playSamples" class="stdlabel">[% "PLUGIN_QOBUZ_PLAY_SAMPLES_DESC" | string %]</label>
 		[% END %]
-
-		[% WRAPPER settingGroup title="PLUGIN_QOBUZ_USE_REPLAY_GAIN" desc="" %]
-			<input type="checkbox" [% IF prefs.pref_useReplayGain %]checked [% END %] class="stdedit" name="pref_useReplayGain" id="pref_useReplayGain" value="1" />
-			<label for="pref_useReplayGain" class="stdlabel">[% "PLUGIN_QOBUZ_USE_REPLAY_GAIN_DESC" | string %]</label>
-		[% END %]
 	[% END %]
 
 [% PROCESS settings/footer.html %]
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -22,7 +22,7 @@ use constant CLICOMMAND => 'qobuzquery';
 use constant MAX_RECENT => 30;
 
 # Keep in sync with Music & Artist Information plugin
-my $WEBLINK_SUPPORTED_UA_RE = qr/\b(?:iPeng|SqueezePad|OrangeSqueeze|Squeeze-Control|Squeezer|OpenSqueeze)\b/i;
+my $WEBLINK_SUPPORTED_UA_RE = qr/\b(?:iPeng|SqueezePad|OrangeSqueeze|OpenSqueeze|Squeeze-Control)\b/i;
 my $WEBBROWSER_UA_RE = qr/\b(?:FireFox|Chrome|Safari)\b/i;
 
 my $GOODIE_URL_PARSER_RE = qr/\.(?:pdf|png|gif|jpg)$/i;
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -41,7 +41,6 @@ $prefs->init({
 	dontImportPurchases => 1,
 	classicalGenres => '',
 	useClassicalEnhancements => 1,
-	useReplayGain => 0,
 });
 
 my $log = Slim::Utils::Log->addLogCategory( {
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -41,6 +41,7 @@ $prefs->init({
 	dontImportPurchases => 1,
 	classicalGenres => '',
 	useClassicalEnhancements => 1,
+	useReplayGain => 0,
 });
 
 my $log = Slim::Utils::Log->addLogCategory( {
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -19,9 +19,10 @@ use Plugins::Qobuz::ProtocolHandler;
 
 use constant CAN_IMPORTER => (Slim::Utils::Versions->compareVersions($::VERSION, '8.0.0') >= 0);
 use constant CLICOMMAND => 'qobuzquery';
+use constant MAX_RECENT => 30;
 
 # Keep in sync with Music & Artist Information plugin
-my $WEBLINK_SUPPORTED_UA_RE = qr/\b(?:iPeng|SqueezePad|OrangeSqueeze|Squeeze-Control)\b/i;
+my $WEBLINK_SUPPORTED_UA_RE = qr/\b(?:iPeng|SqueezePad|OrangeSqueeze|Squeeze-Control|Squeezer|OpenSqueeze)\b/i;
 my $WEBBROWSER_UA_RE = qr/\b(?:FireFox|Chrome|Safari)\b/i;
 
 my $GOODIE_URL_PARSER_RE = qr/\.(?:pdf|png|gif|jpg)$/i;
@@ -125,6 +126,7 @@ sub initPlugin {
 
 	Slim::Control::Request::addDispatch(['qobuz', 'playalbum'], [1, 0, 0, \&cliQobuzPlayAlbum]);
 	Slim::Control::Request::addDispatch(['qobuz', 'addalbum'], [1, 0, 0, \&cliQobuzPlayAlbum]);
+	Slim::Control::Request::addDispatch(['qobuz','recentsearches'],[0, 0, 1, \&_recentSearchesCLI]);
 
 	# "Local Artwork" requires LMS 7.8+, as it's using its imageproxy.
 	if (CAN_IMAGEPROXY) {
@@ -213,18 +215,53 @@ sub handleFeed {
 		items => ( $prefs->get('username') && $prefs->get('password_md5_hash') ) ? [{
 			name  => cstring($client, 'SEARCH'),
 			image => 'html/images/search.png',
-			type => 'search',
+			type => 'link',
 			url  => sub {
 				my ($client, $cb, $params) = @_;
+				my $items = [];
 
-				my $menu = searchMenu($client, {
-					search => lc($params->{search})
-				});
+				my $i = 0;
+				for my $recent ( @{ $prefs->get('qobuz_recent_search') || [] } ) {
+					unshift @$items, {
+						name  => $recent,
+						type  => 'link',
+						url  => sub {
+							my ($client, $cb, $params) = @_;
+							my $menu = searchMenu($client, {
+								search => lc($recent)
+							});
+							$cb->({
+								items => $menu->{items}
+							});
+						},
+						itemActions => {
+							info => {
+								command     => ['qobuz', 'recentsearches'],
+								fixedParams => { deleteMenu => $i++ },
+							},
+						},
+						passthrough => [ { type => 'search' } ],
+		  			};
+	  			}
 
-				$cb->({
-					items => $menu->{items}
-				});
-			}
+				unshift @$items, {
+					name  => cstring($client, 'PLUGIN_QOBUZ_NEW_SEARCH'),
+					type  => 'search',
+					url  => sub {
+						my ($client, $cb, $params) = @_;
+						addRecentSearch($params->{search});
+						my $menu = searchMenu($client, {
+							search => lc($params->{search})
+						});
+						$cb->({
+							items => $menu->{items}
+						});
+					},
+					passthrough => [ { type => 'search' } ],
+	  			};
+
+				$cb->({ items => $items });
+			},
 		},{
 			name => cstring($client, 'PLUGIN_QOBUZ_USERPURCHASES'),
 			url  => \&QobuzUserPurchases,
@@ -949,9 +986,9 @@ sub QobuzGetTracks {
 
 				push @$items, {
 					name  => cstring($client, 'PLUGIN_QOBUZ_ALBUM_NOT_FOUND'),
-					type  => 'text'						
+					type  => 'text'
 				};
-				
+
 				if ($isFavorite) {  # if it's an orphaned favorite, let the user delete it
 					push @$items, {
 						name => cstring($client, 'PLUGIN_QOBUZ_REMOVE_FAVORITE', $albumTitle),
@@ -963,19 +1000,19 @@ sub QobuzGetTracks {
 						nextWindow => 'parent'
 					};
 				}
-				
+
 				$cb->({
 					items => $items,
 				}, @_ );
 			});
 			return;
-			
+
 		} elsif (!$album->{streamable} && !$prefs->get('playSamples')) {  # the album is not streamable
 			push @$items, {
 				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_AVAILABLE'),
-				type  => 'text'						
+				type  => 'text'
 			};
-			
+
 			$cb->({
 				items => $items,
 			}, @_ );
@@ -986,10 +1023,10 @@ sub QobuzGetTracks {
 			my $rDate = _localDate($album->{release_date_stream});
 			push @$items, {
 				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED') . ' (' . $rDate . ')',
-				type  => 'text'						
+				type  => 'text'
 			};
 		}
-		
+
 		my $totalDuration = 0;
 		my $trackNumber = 0;
 		my $works = {};
@@ -1010,7 +1047,7 @@ sub QobuzGetTracks {
 			$totalDuration += $track->{duration};
 			my $formattedTrack = _trackItem($client, $track);
 			my $work = delete $formattedTrack->{work};
-			
+
 			if ( $work ) {
 				# Qobuz sometimes would f... up work names, randomly putting whitespace etc. in names - ignore them
 				my $workId = Slim::Utils::Text::matchCase(Slim::Utils::Text::ignorePunct($work));
@@ -1022,7 +1059,7 @@ sub QobuzGetTracks {
 				# still distinguishing between works with the same name but different composer!
 				$currentComposer = $track->{composer}->{name};
 				if ( $workId eq $lastwork && (!$lastComposer || !$currentComposer || $lastComposer eq $currentComposer) ) {
-					# Stick with the previous value! ($worksWorkId = $worksWorkId;) 
+					# Stick with the previous value! ($worksWorkId = $worksWorkId;)
 				} elsif ( $currentComposer ) {
 					$worksWorkId = $displayWorkId;
 				} else {
@@ -1030,17 +1067,17 @@ sub QobuzGetTracks {
 				}
 
 				# Extended Work ID: will usually not change, but we need to keep non-contiguous tracks from the same work
-				# separate if the user has chosen to integrate playlists with the work titles.				
-				$worksWorkIdE = $worksWorkId;					
+				# separate if the user has chosen to integrate playlists with the work titles.
+				$worksWorkIdE = $worksWorkId;
 				if ( $workPlaylistPos eq "integrated" && $works->{$worksWorkId} ) {
 					if ( $worksWorkId ne $lastWorksWorkId ) {
 						$discontigWorks->{$worksWorkId} = $worksWorkId . $trackNumber;
-					}			
+					}
 					if ( $discontigWorks->{$worksWorkId} ) {
 						$worksWorkIdE = $discontigWorks->{$worksWorkId};
 					}
 				}
-					
+
 				if ( !$works->{$worksWorkIdE} ) {
 					$works->{$worksWorkIdE} = {   # create a new work object
 						index => $trackNumber,		# index of first track in the work
@@ -1048,7 +1085,7 @@ sub QobuzGetTracks {
 						tracks => []
 					} ;
 				}
-				
+
 				# Create new work heading, except when the user has chosen integrated playlists - in that case
 				# the work-playlist headings will be spliced in later.
 				if ( ( $workId ne $lastwork ) || ( $lastComposer && $currentComposer && $lastComposer ne $currentComposer ) ) {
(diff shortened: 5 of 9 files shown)


michaelherger:
If I look at below code I would assume LMS does support track gain out of the box IF there was no `trackGain()` in the `ProtocolHandler.pm` and if `replay_gain` was defined for the track.

https://github.com/Logitech/slimserver/blob/e9c2f88e7ca60b3648b66116240f3f5fe6ca3188/Slim/Player/ReplayGain.pm#L33-53

SamInPgh:
I agree on all points. I really wanted to go further in researching why LMS wasn't handling the `replay_gain` value in the metadata itself but I don't have a functional Unix LMS right now and don't know how to rebuild LMS on my Windows system to do testing there. I sort of figured that, if it could be done via LMS, other streaming services supporting replay gain, such as Pandora and Deezer, would have done it. I would actually much prefer if it was handled by LMS so we could make use of the logic it has to determine when and when not to use track gain, etc.

If we do end up having to use the `trackGain()` callback method, I agree about basing it on the client `replayGainMode` but that could be confusing too if the user selects Album or Smart Gain. Ideally, we can get LMS to handle all that for us.

In the meantime, I guess it's time to unbox the RPi3B I bought last year and install LMS on it.

michaelherger:
Seems like you got a good lead! I sent a mail with some instructions about running LMS on Windows from the source by mail.

darrell-k:
I just added `logError(Data::Dump::dump($track));` where you suggested and got the following output:

```
[23-05-29 18:23:36.7818] Slim::Player::ReplayGain::fetchGainMode (30) Error: bless([
  "-93882434004440",
  1,
  undef,
  undef,
  "qobuz://206247964.flac",
  undef,
  "flc",
  4_517_239,
  "1251.6",
  undef,
  undef,
  undef,
  undef,
  undef,
  "ARCHORA by Iceland Symphony Orchestra from ARCHORA / AI\x{14C}N",
  "ARCHORA BY ICELAND SYMPHONY ORCHESTRA FROM ARCHORA AI\x{14C}N",
  "ARCHORA BY ICELAND SYMPHONY ORCHESTRA FROM ARCHORA AION",
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  0,
  undef,
  undef,
  undef,
  192_000,
  24,
  2,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  undef,
  "e6b859cfecea079b3616a537835f3cf4",
  undef,
  undef,
  undef,
  undef,
  {},
  undef,
  { flc => { init => sub { "???" }, initial_block_type => 1 } },
], "Slim::Schema::RemoteTrack")
```

michaelherger:
> I just added `logError(Data::Dump::dump($track));` where you suggested and got the following output:

Thanks @darrell-k! While I don't know at which position the replay gain would be expected to be, none of those values look like it. The search goes on...

SamInPgh:
Thanks, Darrell! That give me something to go on although, for some reason, I was hoping for a more formatted display. That track has a replay gain of -0.64dB, as you can see from this dump of the metadata taken from the url passed to the protocol handler:

```
[23-05-29 13:51:26.9381] Plugins::Qobuz::ProtocolHandler::trackGain (172) {
  album       => "ARCHORA / AI\x{14C}N",
  albumId     => "b9oaydodijpzb",
  artist      => "Iceland Symphony Orchestra",
  artistId    => 241_887,
  bitrate     => "2386kbps",
  composer    => "Anna Thorvaldsdottir",
  composerId  => 661_663,
  cover       => "https://static.qobuz.com/images/covers/zb/jp/b9oaydodijpzb_600.jpg",
  ct          => "flc",
  duration    => 1251,
  genre       => "Classical",
  genres_list => undef,
  goodies     => undef,
  isClassique => 0,
  performers  => "Iceland Symphony Orchestra, Orchestra, MainArtist - Morten Lindberg, MasteringEngineer - Anna Thorvaldsdottir, Composer, MainArtist - Eva Ollikainen, Conductor, MainArtist - Daniel Shores, MasteringEngineer, MixingEngineer, RecordingEngineer - Ragnhei\xF0ur J\xF3nsd\xF3ttir, Producer, SoundEditor - Joshua Frey, AssistantRecordingEngineer - Collin J. Rae, ExecutiveProducer",
  replay_gain => "-0.64",
  samplerate  => 96,
  samplesize  => 24,
  title       => "ARCHORA",
  type        => "flc",
  version     => undef,
  work        => undef,
  year        => 2023,
}
[23-05-29 13:51:26.9384] Plugins::Qobuz::ProtocolHandler::trackGain (174) Using replaygain value of -0.64 for Qobuz track

```
As you can also see, much of the metadata is not contained in the $track object, apparently including the replay gain as well as composer, release year, etc. That explains the observed behavior. The question is "how does the $track object, which is derived from $song, get populated and how can we get replay_gain in there?".

One possible approach would be to actually populate `$track->replay_gain` from the value returned by Qobuz's protocol handler and then continue on instead of returning the value to the caller. This would cause it to get to the following code:

```
	# only support track gain for remote streams
	if ( $track->remote ) {
		return preventClipping( $track->replay_gain() || $prefs->client($client)->get('remoteReplayGain'), $track->replay_peak() );
	}
```
which would return the correct `replay_gain `value to the caller but without taking into account the `replay_peak` value in the calculation, which we have access to in the plugin but currently don't populate. Note that, for remote tracks (which I assume Qobuz tracks are?), the rest of the logic to determine whether to use smart gain, etc, is bypassed so the net result is going to be the same whether we figure out how to populate `replay_gain `in $track ahead of this point or not.

At this point, I'm leaning towards just going forward with the change to Qobuz but admittedly that is partly due to my not having a usable LMS test environment at the moment, i.e. the path of least resistance. Thoughts?

michaelherger:
> At this point, I'm leaning towards just going forward with the change to Qobuz but admittedly that is partly due to my not having a usable LMS test environment at the moment, i.e. the path of least resistance. Thoughts?

Fine with me. As even if we fixed LMS, we'd need this for backwards compatibility. But please use the global RG settings rather than a plugin specific one to decide whether to use RG or not (`preferences('server')->client($client)->get('replayGainMode')`)

darrell-k:
My general (perhaps naive) thought is that the function is expecting `$track` to be a map and it isn't one. I tried to find out what `$song->currentTrack()` does, and got lost!

SamInPgh:
> > 
> 
> Fine with me. As even if we fixed LMS, we'd need this for backwards compatibility. But please use the global RG settings rather than a plugin specific one to decide whether to use RG or not (`preferences('server')->client($client)->get('replayGainMode')`)

Actually, backward compatibility is another issue that stopped me from contemplating changes to the LMS code, as Pandora and Deezer (as well as other plugins possibly) make use of the protocol handler callback to implement their own replay gain support.

I agree in principle with using the `replayGainMode` prefs option rather than introducing one for the plugin. Their are plusses and minuses to both methods though. The biggest plus for using `replayGainMode` is that it us a per-player option, giving the user a little more flexibility. But do we only bypass it if the option is set to "No volume adjustment"? How about "Album gain" and "Smart gain", which have no real meaning in this context?

One last thing... Since we have access to the peak gain value in the plugin, I'd like to either call (if possible) or duplicate the `preventClipping()` function from the `trackGain()` callback.

```
# Reduce the gain value if necessary to avoid clipping
sub preventClipping {
	my ( $gain, $peak ) = @_;

	if ( defined $peak && defined $gain && $peak > 0 ) {
		my $noclip = -20 * ( log($peak) / log(10) );
		if ( $noclip < $gain ) {
			return $noclip;
		}
	}

	return $gain;
}

```
Whadya think?

michaelherger:
Did you see this posting?... https://forums.slimdevices.com/node/1639950 I dealt with this before, in a different way (for TIDAL). If only my memory didn't fail me all the time 😞.

SamInPgh:
> Did you see this posting?... https://forums.slimdevices.com/node/1639950 I dealt with this before, in a different way (for TIDAL). If only my memory didn't fail me all the time 😞.

Ha ha!! I think your memory is suffering from stack overflow. I'm familiar with the problem.

I just saw that post and am pursuing that approach now. Looks promising! I'll repeat my question from the forum. Do you know if Tidal supports album gain in addition to track gain? If so, this method might not work as well with Qobuz.

darrell-k:
That link is about HTTPS, from today?

SamInPgh:
Michael posted the wrong link. Here is the correct one:

https://forums.slimdevices.com/forum/user-forums/3rd-party-software/94305-qobuz-com-streaming-plugin?p=1639949#post1639949

And the good news is: IT WORKS!!! Not only do we now go through LMS's replay gain logic, we also make use of the `replay_peak `and `remoteReplayGain` values to come up with much more accurate gain values. The whole change now consists of one-line changes in two modules. As a bonus, the Info screen for each track now shows the replay gain value so no more having to load up the Squeeze Ctrl app to see it. There are probably a few other attributes that show up there now too. This made my day! Now I have to figure out how to back out my last commit and replace it with a new one. Hopefully there's an easy way to do that. It would be confusing to have to post another commit basically reversing the last one.

darrell-k:
I don't think it would be too confusing.  The PR will merge the two and only show the net differences. Just explain in the new commit comment.

SamInPgh:
@darrell-k  Can you do another test with ReplayGain.pm and determine whether Qobuz tracks are tagged as "remote"? See line 50. If they are, it will make it very simple to explain to users since currently all remote tracks with any user Replay Gain option other than 'None' are processed as 'Track gain'.  (I know it's late there so tomorrow would be fine.) Thanks.

darrell-k:
Yes:

`track->remote=1`

SamInPgh:
> Yes:
> 
> `track->remote=1`

Thanks! I wonder if it makes any difference when playing imported Qobuz content from the library vs from the plugin. I doubt that it does but this means that none of the Qobuz albums (except those that have been downloaded to the library) can ever take advantage of album gain processing in LMS, should Qobuz ever decide to provide that data. What a shame. It also means that users should now always set a player's Replay Gain option to 'None' when playing Qobuz albums in order to avoid losing the intended dynamic range between tracks. We could always go back later and try to tweak this but at least there is now a way to even the volume out when playing random tracks or playlists.

darrell-k:
Yes, as you've said on the forum, there's no way you would want this switched on when playing an album. I think the functionality is very bad if it means that if album gain is switched on for a player, you get track gain for Qobuz and other remote content. Album gain should mean album gain only!

I tried streaming a Qobuz album from my library view, and as you predicted it's still flagged as remote. 

I've never used replay gain of any kind - my gut feeling has always been not to trust neither the tagging nor the player, and this seems to confirm my prejudice!

darrell-k:
An idea: can we identify the user's current player in ProtocolHandler.pm? If so, how about doing `$song->track->setAttributes($meta);` only if the current player is set to track gain?

SamInPgh:
> An idea: can we identify the user's current player in ProtocolHandler.pm? If so, how about doing `$song->track->setAttributes($meta);` only if the current player is set to track gain?

It's doable but, when it comes to the point where we're trying to trick LMS into doing what we want, it might be time to change LMS (while maintaining backward compatibility, of course). Just a thought.

SamInPgh:
Okay. To take your idea a step further, how about this? In the plugin, while building the tracklist for an album, we compute our own album gain value by feeding all the track gain values into a formula of some sort (in its simplest form, think 'average') and, if the player Replay Gain option is 'Album gain' or 'Smart gain' , we overwrite all the album's track gain values with the calculated album gain value, thus preserving the dynamic gain differences between the tracks while, at the same time, allowing for level differences between albums. The more I think about it, the more I like it. And the best part is that it requires no change to LMS.

On another note, someone (@slartibartfast) on the forum keeps saying that Tidal only provides album gain metadata and not track gain. If this is true, I wonder how that plugin handles the reverse situation when the Replay Gain option is 'Track gain'. I suppose they could approximate it by applying the album gain value to each track.

The more I learn about replay gain, though, the less inclined I am to use it myself except maybe for a random party mix.

darrell-k:
I think we'd need to use the lowest track gain in the album in order to avoid potential clipping (I'm guessing that the preventClipping function acts on individual tracks, so it could put the gain out of alignment if we rely on that).

We'd also need to do this when building a track list from a work playlist rather than the whole album (consistent replay gain is particularly important in this case).

SamInPgh:
Well, I bit the bullet and reactivated my Tidal sub. Check out the results on the forum. Looks like Tidal agrees with you, Darrell. They seem to take the lowest track gain for the tracks on an album and apply it to all the tracks. In my example album, the highest track gain from Qobuz was +2.41dB (for the spoken intro) and the lowest was -5.81dB. Tidal used -6.03dB for every track, giving them a 0.22dB cushion. By making the change to the plugin that I propose, Qobuz would be the first streaming service for which LMS provides both track gain and album gain support. However, I would prefer to make the change in LMS by  modifying the logic in ReplayGain.pm that applies the track replay_gain value for all remote tracks when the Replay Gain option is not set to 'None'. That is just wrong. If we allowed the logic flow to proceed past that point for remote tracks, we could make a simpler change in the Qobuz plugin to simply insert the lowest track gain value into the replay_gain tag of the album item associated with the tracks, which is where it belongs, allowing the 'Album gain' (and possibly 'Smart gain') options to function properly.

darrell-k:
Great investigations! I agree with the objective, the only thing that concerns me is that unlike plug-in updates, which are semi-automatic or even automatic with the preference set, LMS updates require manual install steps in the OS, needing admin/root access. I was running a years out of date LMS until relatively recently, when something (I can't remember what) just stopped working. 

I'm not saying we shouldn't modify LMS, but could we also "preview" the feature through the plug-in route?

SamInPgh:
I think it's clear that we could get it at least mostly working the way we want by doing all the changes in the plugin. However, the more I look at ReplayGain.pm, the more I think it needs to be changed ASAP. For example take this now-infamous statement:

```
	# only support track gain for remote streams
	if ( $track->remote ) {
		return preventClipping( $track->replay_gain() || $prefs->client($client)->get('remoteReplayGain'), $track->replay_peak() );
	}

```
Picture a user sitting back for a relaxing listening session.  They are playing a random playlist of songs from their library, some of which are remote tracks (Tidal, Qobuz, etc.) and some which are not.  The player's Replay Gain is set to Track Gain, which seems reasonable in this case. Let's say the user also has `remoteReplayGain` set to a high positive or negative value for whatever reason, say +15 (the UI allows for values between -20 and +20). When a remote track with a non-zero `replay_gain` is processed -- let's say it's -5.1dB -- the above line of code will pass that value as the $gain parameter to the `preventClipping()` routine along with the track's `replay_peak` value, if any. Let's say the next track coming in is also a remote track with a `replay_gain `value of 0, which is not uncommon. The way I read that line of code, it will pass the remoteReplayGain value -- in this case +15dB -- to `preventClipping()` as the $gain (again along with the track's `replay_peak` value, which makes no sense), probably resulting in our user jumping out of his/her chair, stepping on the dog, and possibly suffering heart failure, depending on the track. Maybe I'm reading it wrong or maybe I'm just missing the logic, but this again seems very wrong to me. For one thing, I feel that the `remoteReplayGain` setting should be applied evenly to all tracks, whether they have a non-zero `replay_gain` value or not. If nothing else just add the values together instead of logical OR'ing them. Going further though, the `remoteReplayGain` value should be applied to ALL remote tracks under all circumstances, not just when `replay_gain` is being used. (Anybody looking for a slightly-use soapbox?)

I'm going to be very embarrassed if I'm completely off base here but that's okay -- I can take it. Give me some feedback on this. Thanks.

michaelherger:
I think your reading is valid. I tend to believe that listeners suffer heart failure, as we'd have heard about the dog incidents had the listeners survived.

The most important point in your analysis though is probably the `replay_gain == 0` vs. unknown value. We currently have no way to know whether a handler set that value to zero, or whether it can't deal with it at all. The best might be to check whether `defined $track->replay_gain` was true before falling back. Or should we introduce something like a `canHandleGain` in the protocol handlers to skip `remoteReplayGain` in those cases?

SamInPgh:
> I think your reading is valid. I tend to believe that listeners suffer heart failure, as we'd have heard about the dog incidents had the listeners survived.
> 
LOL!!! 😲

darrell-k:
How about passing the lowest replay gain specified for the group of tracks being added, album or work (or the individual gain value when adding a single track), but if that turns out to be zero, pass -0.1?

We could then move on to fixing the LMS code.

darrell-k:
I've done some testing with this PR installed and my remote replay gain preference set to 20:

With non-zero `replaygain_track_gain` from Qobuz, it works as expected:
```
[23-05-31 12:59:03.6800] Slim::Player::ReplayGain::fetchGainMode (59) Error: remote replay gain pref=20
[23-05-31 12:59:03.6802] Slim::Player::ReplayGain::fetchGainMode (60) Error: track gain=0.2
[23-05-31 12:59:03.6803] Slim::Player::ReplayGain::fetchGainMode (61) Error: track peak=0.870972
[23-05-31 12:59:03.6806] Slim::Player::ReplayGain::fetchGainMode (63) Error: return="0.2"
```

When forcing track_gain to zero, as long as `replaygain_track_peak` is non-zero, preventClipping does its job and something vaguely sensible is returned:
```
[23-05-31 13:00:38.0336] Slim::Player::ReplayGain::fetchGainMode (59) Error: remote replay gain pref=20
[23-05-31 13:00:38.0338] Slim::Player::ReplayGain::fetchGainMode (60) Error: track gain=0
[23-05-31 13:00:38.0340] Slim::Player::ReplayGain::fetchGainMode (61) Error: track peak=0.870972
[23-05-31 13:00:38.0342] Slim::Player::ReplayGain::fetchGainMode (63) Error: return="1.19991612926017"
```

But if `replaygain_track_peak` is also zero, all bets are off:
```
[23-05-31 13:01:54.1734] Slim::Player::ReplayGain::fetchGainMode (59) Error: remote replay gain pref=20
[23-05-31 13:01:54.1736] Slim::Player::ReplayGain::fetchGainMode (60) Error: track gain=0
[23-05-31 13:01:54.1738] Slim::Player::ReplayGain::fetchGainMode (61) Error: track peak=0
[23-05-31 13:01:54.1740] Slim::Player::ReplayGain::fetchGainMode (63) Error: return=20
```
So it is likely that the number of flat dogs may be less than we feared, but it may still be greater than zero.

SamInPgh:
So, as far as Qobuz goes, now that we populate `replay_peak` in the track attributes, we should be okay in terms of canine protection (okay, that analogy is now officially retired) and could make the change to populate all the track gain values for an album (or work) with the lowest one of the group without worrying about that value being zero as long as we pass a non-zero replay_peak value too  However, I don't know that Tidal or any other streaming services populate `replay_peak` at all. Unfortunately for me (but maybe fortunately for the LMS community), I don't yet have a working test version of LMS but I'd like to experiment with allowing remote tracks to get past the `if $track->remote` roadblock. If you look at some of the code further down that processes Smart Gain, there are places where it is checking for a remote track but, as we now know, it will never see one.

Meanwhile, I'll start looking at what it would take to implement the strategy of populating the RG of all tracks in an album/work with the lowest value of the group (along with its corresponding `replay_peak` value) if the player is using Album or Smart Gain.

SamInPgh:
Just an update. Now that I know what `ReplayGain.pm` actually does with remote tracks (nothing useful), I am going back to Plan A, which is to implement the `trackGain()` callback in `ProtocolHandler.pm` so that we have complete control of the gain for each track. If the player is set to use Album Gain or Smart Gain, we will pass back the album gain based on the lowest track gain (without using `remoteReplayGain` in the calculation). If, in the future, the logic in `ReplayGain.pm` is changed in a way that makes more sense for Qobuz, it would be a simple change it back again, now that the RG attributes are being passed to it.

I'm not going to rush this as we've been living without RG in Qobuz for awhile, so don't wait for me if you want to push out a new release with the recent searches feature.

SamInPgh (commented):
This isn't going to work. I need to figure out how to wait for the async API calls in `trackGain()` to complete before returning. The way to do this is eluding me. Any advice would be appreciated.

michaelherger:
You can't wait for an async call. It's not an option. You'd have to do this "out of band" somehow. A good candidate would be `getNextTrack()`: it's an async call already. It's called whenever a new track is due to be played. Perfectly placed to not only get information about the track itself, but about the album, too. Store the result of the album metadata lookup in the cache. And when `trackInfo()` is called you read from the cache (which is sync). 

Luckily enough `getAlbum()` would already store additional album information for each track. Add storing of the gain value there. You should then be able to grab all cached track information using `my $meta = $cache->get('trackInfo_' . $trackId);`, or using `getTrackInfo()` in a sync way for additional sanity checking:

```perl
my $meta = Plugins::Qobuz::API->getTrackInfo(sub {}, $trackId);
```

Neither call is guaranteed to return the data as a cache can be incomplete/expired. But it's probably your best bet at this point.

SamInPgh:
Thanks to you both for your observations and suggestions. They are very helpful  I am closing in on a way to eliminate async calls in the callback which, in my opinion, is mandatory. I have a question though. Are the keys that can be used to access the cache listed or documented anywhere? I'm referring to, e.g., "trackId_$trackId" and "album_with_tracks_$albumId".

SamInPgh:
> Documentation?... Obviously no. But there's that `precache` stuff which should give you an idea.

I need more than an idea. I need to know how to access an album object without the tracks. The only examples I have from the code are the two I listed. There must surely be others.

SamInPgh:
No, it wouldn't. When an album is played directly from a list of albums, without first exploding the track list, the album is apparently cached but not the tracks. At that point, a call to `$cache->get('album_with_tracks_' . $albumId)` returns null even though a `_get()` of the same $albumId via `GetAlbum() ` returns the album item from the cache. However, if the album is first exploded and then played, the call to `$cache->get('album_with_tracks_' . $albumId)`  returns a pointer to the album and its tracks. This issue is the key (no pun intended) to getting this feature working in an efficient and predictable way. I suppose I could add another API function such as `isAlbumCached()` but it would be much easier if there were already a predefined key similar to  `trackId_$trackId` that I could use with `$cache->get()`.

SamInPgh:
Never mind. I created my own and it works marvelously. It's similar to `trackInfo_$trackId` in that it stores a subset of metadata. It is called` albumInfo_$albumId`. It is written in `_preCacheAlbum `and used by the `trackGain `callback. I'm now ever-so-close to getting this thing working very well. (Have you heard that before?). More to come...

SamInPgh:
Following copied from my PM thread with Darrell that I mistakenly thought @michaelherger was included in:

"Actually, the whole thing has been re-designed and re-engineered. There is no use of async functions in the callback and it is much more straightforward due to the addition of caching some album metadata (including the calculated album replay gain values) that the callback makes use of. Fyi, the logic makes no attempt to implement Smart Gain, treating it as equivalent to Album Gain."
 
"I just took another look at the LMS ReplayGain logic and found that, if either Album Gain or Smart Gain is in use and the current track doesn't have an album associated with it, it returns a replay gain value of 0 for that track. In other words, Smart Gain isn't really that smart. I have a proposal. I'd like to make the plugin better than LMS in that regard. It's a simple change. Instead of applying track gain for these tracks when Album Gain is in use, we do as LMS does and return a value of 0, only using the track gain for Smart Gain. In effect, then, we would be doing what LMS says it's doing but really isn't and we would also have clearly defined differences among all flavors of Replay Gain in the plugin. It's basically a one-line change. Thoughts?"
 
"That was more confusing than I intended. Let me try again. With the Album Gain option in effect, individual tracks not associated with an album (e.g. from a playlist or LMS favorite) would be assigned a replay gain of 0. With the Smart Gain option in effect, however, these tracks would be assigned the track's replay gain value, thus conforming to the definition of Smart Gain and giving the user a clear differentiation between those two options. The change is ready to commit when we have agreement. (My vote is 'Yes').

michaelherger:
I don't really have an opinion about SmartGain, as I'm not using ReplayGain at all... I'm mostly an album listener. I do adjust volume every few hours manually if needed 😀.

As for the caches: I think I've introduced a bit of a mess there myself. I'm sure this could be optimised. But then... oh well, if it works and there's nobody complaining about waste of disk space? Your solution sounds ok to me.

darrell-k:
My reading of what LMS *says* it does is that if album gain is set in the preferences, the album gain value is always used, which would be my preference.

darrell-k:
I might have found an inconsistency with preference album gain: as expected, playing a track directly, without going through the album, uses track gain. Then playing the same track via the album uses album gain. All OK so far.

But then clear the playlist and then play the song directly (not through the album) again: it still uses album gain.

I'm guessing that adding the track via the album caches the album gain, which we're then stuck with until the cache entry expires.

SamInPgh:
> My reading of what LMS _says_ it does is that if album gain is set in the preferences, the album gain value is always used, which would be my preference.

What it doesn't say is that the album gain is not always (ever?) available for individual tracks and, in that case, no replay gain is applied. My feeling is that you would find a lot of interesting "inconsistencies" with replay gain as calculated in `replayGain.pm`. 

I am going to make the change to apply no replay gain if Album Gain is in effect and no album gain is available. Smart Gain will continue to use the track gain in that scenario.

As for the inconsistent application of album gain to individual tracks, that could easily be addressed by changing the `albumInfo` cache expiration to be the duration of the album or some such shorter interval, which I was considering doing so we wouldn't be "stuck with" the cached album gain for 30 days. In fact, on my first quick reading of the following line in `_preCacheTrack`, I thought this was being done there:

`	$cache->set('trackInfo_' . $track->{id}, $meta, ($meta->{duration} ? QOBUZ_DEFAULT_EXPIRY : QOBUZ_EDITORIAL_EXPIRY));`

Upon closer reading, however, l realized that, if the track has a non-zero duration (don't they all?), the expiration is set for 30 days, the logic of which I don't actually understand. In any case, however, the following statement from the LMS Audio options menu applies equally to the Album Gain option, whether applied to LMS or Qobuz:

 _"Album gain" can be used to make sure that all albums sound equally loud, but that volume differences between tracks within an album are preserved._

Although no mention is made about the treatment of individual non-album tracks in a playlist, the following is also true in both scenarios:

_Individual tracks in a playlist will use album gain, if available. Otherwise, no gain is applied._

Let me know what you think about expiring the `albumInfo` cache entries earlier than 30 days and what interval might make the most sense. I'm leaning towards using the album duration.

SamInPgh:
I have made a change such that non-album tracks played in Album Gain mode will NEVER have `album gain` (or any gain) applied, even if it is available via the cached album info. These same tracks will ALWAYS have `track gain` applied when running in Smart Gain mode.  All other situations are straightforward according to the Replay Gain mode setting, and every situation should now be completely consistent and predictable, regardless of whether album info has been cached, etc.  

_My reading of what LMS says it does is that if album gain is set in the preferences, the album gain value is always used, which would be my preference._

First of all, I don't believe that is what LMS actually does. Secondly, why would one want the album gain value applied to non-album tracks in this scenario? For that matter, why would one be using Album Gain when listening to a playlist containing a mixture of albums and individual tracks? Smart Gain would probably be a better choice. in that the individual tracks would at least have track gain applied.  As for me, after this PR is merged, I will probably go back to never using any Replay Gain option, except maybe when using Random Mix or playing Qobuz playlists, where I might use Track Gain. When possible,, I try to preserve the "bit perfect"-ness of the original material throughout the playback chain by using the following Player settings:

Volume Control:   Output level is fixed at 100%
Replay Gain:   None
Default Adjustment For Remote Streams:   0

darrell-k:
I use exactly the same settings as you, and likewise don't use replay gain of any flavour - not least because I was never sure what it would actually do, suspicions only confirmed by your investigations here. 

I was trying to put myself in the head of someone who did listen to a mixture of tracks and whole albums and who wanted some level of volume equalisation while preserving album dynamic range. If that were me, I'd want a "set and forget" option which I think would be a single gain level for each album, whether playing all together or in a playlist. This would mean that any given track would always play at the same relative volume. Unless an album was mastered by a complete lunatic, it would be a pretty good equalisation of average volume with least manipulation, if that makes any sense. 

I suppose the shorter way to say what I'm trying to say is that I would *never* use an option which applied track gain.

Having said all that, I think what you're proposing is reasonable - but there will be questions from users!

SamInPgh:
Duly noted, and I stand ready to answer any and all questions. At this point, I consider the feature ready for rollout. Many thanks to both of you for your helpful suggestions and constructive criticism over the course of this process. I can only shudder to think of the version(s) I was ready to release earlier. Thank god for the code review process!

darrell-k:
And the questions keep coming...

What is the purpose of the added `|| !defined $meta->{goodies}` ?

SamInPgh:
It's a shortcut for detecting a track that wasn't added to the playlist as part of an album so that its replay gain value can be set correctly for both  the `Track gain` (0) and `Smart gain` (track gain) option settings. Have you found any inconsistencies in your testing? There are a lot of different scenarios that can occur when using mixed playlists.

darrell-k:
That's what I thought. I was worried about trusting Qobuz to always return a goodies tag, although it does seem to return an empty list when there are none.

But in testing this out, I may have come across a wider problem. Stay tuned...

SamInPgh:
Wider than what?  😁

darrell-k:
I was hoping to come up with a solution rather than just keep reporting problems, but I have fried my brain so I'll lay out the issue here, while continuing to try to solve it - two heads are better than one, hopefully.

Firstly, it might be best to map out the expected behaviour, in case I'm not understanding it right. Is this correct?

1. Track Gain
Source of track: album listing - use: track_gain
Source of track: songs listing - use: track_gain
Source of track: playlist (local, user-Qobuz, Qobuz-curated) - use: track_gain

2. Album gain
Source of track: album listing - use: computed album_gain
Source of track: songs listing - use: zero gain
Source of track: playlist  (local, user-Qobuz, Qobuz-curated) - use: zero gain

2. Smart gain
Source of track: album listing - use: computed album_gain
Source of track: songs listing - use: track_gain
Source of track: playlist (local, user-Qobuz, Qobuz-curated) - use: track_gain

So, assuming this is a correct understanding, the problem (using album gain) is this:

Play a track from a Qobuz album listing, then save it to a local (SBS) playlist. When played from this created playlist, the album gain will be applied (we would expect gain not to be used, which it won't be if you add the track from a Qobuz Song item). In fact, playing from the playlist will result in different gains being applied depending on the context in which you last played the song from a Qobuz menu - see below for steps to recreate this.

1. Add a song to the play queue via a Qobuz songs display
2. Play - replay gain is set to zero - CORRECT
3. Save the play queue as a local playlist
4. Go to the created SBS playlist
5. Add the track to the play queue
6. Play - replay gain is set to zero - CORRECT
7. Go to the same song in Qobuz, but this time via its album listing
8. Add the track to the play queue
9. Play - replay gain is set to computed album_gain - CORRECT
10. Go to the created SBS playlist
11. Add the track to the play queue
12. play - replay gain is set to computed album_gain **INCORRECT**

Note that this does not seem to be a problem with playlists on Qobuz - only when Qobuz tracks are played from a local SBS playlist.

I think what we really need to know in the trackGain subroutine is where the play queue item originated from. But how do we do that?

SamInPgh:
I think we're getting way more into the weeds than is necessary at this point. First of all, if we subjected the LMS Replay Gain processing to this level of rigorous testing (which, by the way, we can't because the author wisely omitted any debug logging in the module whatsoever, leaving only this not-very-useful Info logging statement) : 	

`main::INFOLOG && $synclog->info( "ReplayGain: track=$track" );`

we would spend the rest of our lives trying to get it to work consistently and reliably.

I understand what you're saying and yes, replay gain can be applied in a slightly different manner when using Album Gain, depending on the source of the track and in the context of a playlist, etc. However, in the vast majority of cases, which would be playing albums or album content directly from the Qobuz library or from the LMS library, and in ALL cases when using Track Gain or Smart Gain, it will work both consistently and reliably. Over time, we may come up with ways to improve Album Gain processing, as well as other aspects of the feature and, in the long run, what I'd really like to do is to enhance and improve the LMS code to handle subscription-based streaming tracks so we don't have to do it at all in the plugin. As I said, I will gladly take ownership responsibility for this feature and handle any questions from users while it is in its initial rollout phase, but let's not prevent a major enhancement like this from being released because it may not be perfect. Remember that this is an optional feature and the user has complete control over when and how to use it (or not use it, as is the case with you, Michael, and myself).  ;-)

SamInPgh:
Fyi, I just did a test on the only local album I could find in my library that has replay gain tags, although there may be others. (It sure would be nice if Advanced Search supported it as a search argument.) The album has replay gain tags for each track but no album gain, which I think is fairly common. Here is what happens when playing a track from this album with different Replay Gain settings:

None:  No RG applied
Track:   Track RG applied
Album: No RG applied
Smart:  Track RG applied

So it matches the behavior of the plugin when a Qobuz track is played with no album gain available for it, i.e. no gain applied for Album Gain; track gain applied for Smart Gain.

Darrell, do you have any local albums that have album gain tags only? How about ones that have both? Is it usually only one or the other? I have a feeling that RG tagging is generally more common on classical titles than other genres. It would be helpful to know what LMS does just for comparison. I'm not concerned if we deviate a little in certain less common situations such as local playlists vs Qobuz playlists, etc. As I said, this is just v1.0 of RG support in Qobuz and it's already way ahead of Tidal or Spotify in that regard.

darrell-k:
I used metaflac a little while ago to add replay gain tags to a local album for testing:
`metaflac --add-replay-gain /path/to/album/dir/*.flac`
This adds track and album gain:
You could use `metaflac --remove-tag=...` to remove either album or track gain tags.

Windows version from https://xiph.org/flac/download.html if you don't have it.

SamInPgh:
Thanks. It will come in handy for testing if and when it is decided to add replay gain support for  subscription streaming services in LMS. In the meantime, I think we're ready for "prime time"!

michaelherger:
Here are two thoughts that came to my mind when I tried to digest your numerous comments (I'm sorry, I've been out for a few days, seeing more music on stage than I've probably done in a decade...):

* You seem to have done some serious work on the remote replay gain aspect, discovering many limitations of LMS' existing code. It almost feels as if we should improve this in the core code.
* You're not even interested in the whole thing yourself - what a pity! I would have loved to see a patch for LMS 😁.
* Ok, one more bonus thought: does any replay gain setting other than off and smart even make sense? Shouldn't we always apply smart gain?

As I've mentioned before I'm not using this myself. So most of my understanding (without looking a the code) is rather superficial. But I'd confirm @darrell-k's listing of cases. Except that playlist and "song listing" probably are the same, unless all the songs in that list happen to be from the same album. We should not distinguish based on where tracks come from, but decide at play time whether all tracks are from the same album or not. And that's what I thought what Smart Gain was doing. Am I wrong there?

darrell-k:
It's not quite that simple. Testing smart gain with local albums:

Add one track - plays with track gain
Add the next (contiguous) track - plays with album gain
Skip back to the first track - plays with album gain
Add a non-contiguous track (or a track from a different album) - this one will play with track grain
Skip back to one of the contiguous tracks - will play at album gain.
Remove one of the contiguous tracks - everything now plays at track gain.
etc

So contiguous tracks from the same album will play at album gain, regardless of what else is in the play queue. Everything else in the play queue will play at track gain.

darrell-k:
On your bonus thought: given how smart gain is working now,  track gain could still be useful for mixed playlists, and album gain would still bring tracks from different albums closer in relative volume while maintaining album dynamic range even across non-contiguous tracks - a use case for this might be a user who removes a couple of tracks that they really don't like while playing an album.

SamInPgh:
Darrell's description of how LMS handles Smart Gain with local content is correct, and is the only difference between how it and Qobuz process the Replay Gain setting. In Qobuz, Smart Gain isn't as smart. It doesn't do any analysis of the previous and next tracks to make the determination between album gain and track gain. It simply uses album gain if it is available for the current track and, if it is not, it uses track gain which, for Qobuz, is always available.  It is up to the user to employ the Replay Gain setting tool in the proper way to achieve what they want to do. For instance, when playing a random mix of tracks, such as one of the Qobuz playlists, Track Gain would be a good choice if the listener wanted to normalize the volume of all the tracks to the theoretical peak gain of 89dB, which is what all these gain values are supposed to do. In this case, Smart Gain would achieve the same goal because the tracks in these Qobuz playlists would not contain album gain values.  When playing a Qobuz album, however, using Track Gain would only be useful if, again, one wanted to normalize all the album tracks to 89dB, which would not be desirable for most listeners. However, it is up to the user to set Replay Gain according to what listening experience they desire. Fortunately, it can be set separately for each player.  In my case, I will almost always use "None" for my main listening room, where I desire to get the signal to my DAC with a minimum of processing from LMS --- ideally none. However, for the Google Nest Mini pair in my bedroom, where I often listen to mixes or playlists and have no illusions about bit-perfection, I will probably just leave it set to Smart Gain.

Are there any objections to getting this released to the users? For many of them, Replay Gain is an important addition and let me say once more that having support for both track gain and album gain is a groundbreaking event for LMS streaming services. Tidal supports only album gain and I believe Deezer supports only track gain. It's basically a selling point for Qobuz among LMS users. Maybe we should ask them for royalties.  ;-)

SamInPgh:
P.S.  @michaelherger How was the Peter Gabriel concert? I'm envious! What other acts did you see on your holiday?

SamInPgh:
Okay. I was going to save this for a future release but you've given me too much time now and I'm bored. I'm going to implement support for the `remoteReplayGain` preference setting for Qobuz in the plugin callback. In LMS (replayGain.pm), it is (was) applied only if the Replay Gain option was not set to "None" AND the replay gain value was zero (recall the flat dog scenario). It was then applied INSTEAD OF track gain. I am considering the following options:

1. Only apply it if Replay Gain is not set to "None" and both the `replay_gain `and `replay_peak` values are zero. So essentially the same behavior as before but with an extra check to prevent flat dogs. (This would basically never happen because just about every track in Qobuz has a track gain value as far as I can tell and, even if happens to be zero, the peak value would be non-zero.)

2. Same as option 1, but always apply the `remoteReplayGain `by adding it to the track or album gain and not replacing it.

3. Same as opton 2, but apply `remoteReplayGain` even if the Replay Gain option is set to "None".

Options 2 and 3 allow the user the option of specifying and using `remoteReplayGain` and track/album replay gain independently of each other, yielding more flexibility, while option 3 adds the ability to fine tune the gain level of Qobuz content without having to use Replay Gain at all, i.e. "None".

My preference is option 3 for the maximum flexibility while retaining the ability for "bit perfect" play by using Replay Gain of "None" and `remoteReplayGain` of 0. 

Your thoughts please, including other alternative options. Thanks.
`

SamInPgh:
Never mind. After doing some testing, the average gain levels between local tracks and Qobuz is actually minimal and `remoteReplayGain` can still be used to balance other remote streaming sources if needed. Therefore, I am officially done with this PR. Over and out.

No, really...

darrell-k:
I agree. It's a real step forward as it is.

The only thing I'd suggest is adding a comment to explain the use of
$meta->{goodies}, to guard against it being populated outside of the album
listing context in the future, which would screw up your decision logic.

Or even better, if not too much trouble, add a new element to $meta
(isAlbumTrack ?) which would be more descriptive and hopefully safer (I
wouldn't exactly trust Qobuz not to change their use of the goodies tag
arbitrarily).

SamInPgh:
Done. New `isAlbumTrack `item added to trackInfo metadata, populated in `_precacheAlbum`, and checked by the callback to determine whether or not the track is being played from an album.

> I agree. It's a real step forward as it is. The only thing I'd suggest is adding a comment to explain the use of $meta->{goodies}, to guard against it being populated outside of the album listing context in the future, which would screw up your decision logic. Or even better, if not too much trouble, add a new element to $meta (isAlbumTrack ?) which would be more descriptive and hopefully safer (I wouldn't exactly trust Qobuz not to change their use of the goodies tag arbitrarily).

SamInPgh:
Okay. That didn't work well. The change worked fine for Qobuz content played from the plugin view but the new `isAlbumTrack` element wasn't being picked up for albums in the LMS library even after a rescan. I'm sure there is a simple explanation but I reverted back to checking a different $meta element that is more reliable as, if Qobuz changed it, we would have more to worry about than this. The line of code in `trackGain` has also been commented to make it clear what it is doing.

michaelherger:
Thanks a lot for all your work and investigations. I learned (refreshed) a bit about that topic usually ignore. Let's get this out to folks, so it gets tested.

PR #44, opened 2023-05-29
Add search history to the Qobuz search facility

darrell-k:
Code mostly stolen from the BBCSounds plug-in, but any bugs are mine alone!

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -5,6 +5,7 @@ use base qw(Slim::Plugin::OPMLBased);
 
 use JSON::XS::VersionOneAndTwo;
 use Tie::RegexpHash;
+use POSIX qw(strftime);
 
 use Slim::Formats::RemoteMetadata;
 use Slim::Player::ProtocolHandlers;
@@ -982,8 +983,9 @@ sub QobuzGetTracks {
 		}
 
 		if (!_isReleased($album) ) {
+			my $rDate = _localDate($album->{release_date_stream});
 			push @$items, {
-				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED') . ' (' . $album->{release_date_stream} . ')',
+				name  => cstring($client, 'PLUGIN_QOBUZ_NOT_RELEASED') . ' (' . $rDate . ')',
 				type  => 'text'						
 			};
 		}
@@ -1209,8 +1211,9 @@ sub QobuzGetTracks {
 				push @$items, $item;
 			}
 
+			my $rDate = _localDate($album->{release_date_stream});
 			push @$items, {
-				name  => cstring($client, 'PLUGIN_QOBUZ_RELEASED_AT') . cstring($client, 'COLON') . ' ' . $album->{release_date_stream},
+				name  => cstring($client, 'PLUGIN_QOBUZ_RELEASED_AT') . cstring($client, 'COLON') . ' ' . $rDate,
 				type  => 'text'
 			};
 
@@ -1294,7 +1297,7 @@ sub _albumItem {
 	my $artist = $album->{artist}->{name} || '';
 	my $albumName = $album->{title} || '';
 	my $showYearWithAlbum = $prefs->get('showYearWithAlbum');
-	my $albumYear = $showYearWithAlbum ? $album->{year} || (localtime($album->{released_at}))[5] + 1900 || 0 : 0;
+	my $albumYear = $showYearWithAlbum ? $album->{year} || substr($album->{release_date_stream},0,4) || 0 : 0;
 
 	if ( $album->{hires_streamable} && $albumName !~ /hi.?res|bits|khz/i && $prefs->get('labelHiResAlbums') && Plugins::Qobuz::API::Common->getStreamingFormat($album) eq 'flac' ) {
 		$albumName .= ' (' . cstring($client, 'PLUGIN_QOBUZ_HIRES') . ')';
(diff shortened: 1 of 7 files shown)


Plugin.pm
```diff
@@ -1800,4 +1837,87 @@ sub _localDate {  # convert input date string in format YYYY-MM-DD to localized
...
+	my $search = shift;
+	main::DEBUGLOG && $log->is_debug && $log->debug("++addRecentSearch");
+
+	my $list = $prefs->get('qobuz_recent_search') || [];
+
+	$list = [ grep { $_ ne $search } @$list ];
+
+	push @$list, $search;
+
+	$list = [ @$list[(-1 * MAX_RECENT)..-1] ] if scalar @$list > MAX_RECENT;
```
michaelherger, line 1850: You could rewrite this as `$list = [ splice(@$list, MAX_RECENT) ] if ...`

Plugin.pm
```diff
@@ -1800,4 +1837,87 @@ sub _localDate {  # convert input date string in format YYYY-MM-DD to localized
...
+					cmd    => ['qobuz', 'recentsearches' ],
+					params => {
+						delete => $del
+					},
+				},
+			},
+			nextWindow => 'parent',
+		  },
+		  {
+			text => 'Clear search history',
```
michaelherger, line 1895: Please use a string token here.

Plugin.pm
```diff
@@ -1800,4 +1837,87 @@ sub _localDate {  # convert input date string in format YYYY-MM-DD to localized
...
+			},
+			nextWindow => 'grandParent',
+		  };
+
+		$request->addResult('offset', 0);
+		$request->addResult('count', scalar @$items);
+		$request->addResult('item_loop', $items);
+	}elsif ($request->getParam('deleteAll')) {
+		$prefs->set( 'qobuz_recent_search', [] );
+	}elsif (defined $request->getParam('delete')) {
```
michaelherger, line 1913: Please add a space in front of the `elsif`.

Plugin.pm
```diff
@@ -1800,4 +1837,87 @@ sub _localDate {  # convert input date string in format YYYY-MM-DD to localized
...
+				go => {
+					player => 0,
+					cmd    => ['qobuz', 'recentsearches' ],
+					params => {
+						deleteAll => 1
+					},
+				}
+			},
+			nextWindow => 'grandParent',
+		  };
```
michaelherger, line 1906: Please remove the two leading spaces.

Plugin.pm
```diff
@@ -213,18 +215,53 @@ sub handleFeed {
...
+								command     => ['qobuz', 'recentsearches'],
+								fixedParams => { deleteMenu => $i++ },
+							},
+						},
+						passthrough => [ { type => 'search' } ],
+		  			};
+	  			}
+
+				unshift @$items, {
+					name  => 'New Search',
```
michaelherger, line 248: Please use string token for the localization.

michaelherger (changes requested):
Mostly looking fine. But please use string tokens to make the menu localizable. Thanks!

Plugin.pm
darrell-k: Where are the system-wide translations? (I wonder if there are existing ones I can use)

michaelherger: System wide are in LMS code in the root's `strings.txt`. If you can't find anything suitable there you can add your own.

darrell-k: Nothing suitable in there, so I should add to the plugin's  strings.txt?

michaelherger: Sure.

Plugin.pm
darrell-k: This didn't work - the array is populated oldest to newest (and then presented in reverse order) and we want to remove the oldest elements over the limit. a `shift` did the job.

darrell-k:
All done!

strings.txt
```diff
@@ -403,3 +403,17 @@ PLUGIN_QOBUZ_ALBUM_NOT_FOUND
 	DE	Album nicht in der Qobuz-Bibliothek gefunden
 	NL	Album niet gevonden in de Qobuz bibliotheek
 	FR	Album introuvable dans la bibliothèque Qobuz
+	
+QOBUZ_CLEAR_SEARCH_HISTORY
```
michaelherger, line 407: I'm sorry to be a pain... but by convention we should be using `PLUGIN_QOBUZ_...`.

Plugin.pm
```diff
@@ -1800,4 +1837,87 @@ sub _localDate {  # convert input date string in format YYYY-MM-DD to localized
...
+	my $search = shift;
+	main::DEBUGLOG && $log->is_debug && $log->debug("++addRecentSearch");
+
+	my $list = $prefs->get('qobuz_recent_search') || [];
+
+	$list = [ grep { $_ ne $search } @$list ];
+
+	push @$list, $search;
+
+	shift(@$list) while scalar @$list > MAX_RECENT;
```
michaelherger, line 1850: `splice(@$list, -1 * MAX_RECENT - 1)`?

Plugin.pm
darrell-k: Yes, I have that page in my bookmarks - it's where I found the solution: `shift(@a) == splice(@a,0,1)`, given that in use we're never going to have more than MAX_RECENT + 1 elements in the array. I only replaced the `if` with `while` to cater for the extremely unlikely case where MAX_RECENT is made smaller in a future release of the plug-in.

Plugin.pm
darrell-k: I think `shift(@list) while...` reads better. See also reply above.

PR updated with the strings.txt change.

michaelherger: Fair enough!

michaelherger:
Thanks!

PR #47, opened 2023-06-19
New preferences: show disc/track number and show Parental Warning flag

darrell-k:


Plugin.pm
```diff
@@ -1365,6 +1368,11 @@ sub _albumItem {
 		$item->{line2} = $artist . ($albumYear ? ' (' . $albumYear . ')' : '');
 		$item->{name} .= $albumYear ? "\n(" . $albumYear . ')' : '';
 	}
+	
+	if ( $prefs->get('parentalWarning') && $album->{parental_warning} ) {
+		$item->{name} .= ' [E]';
+		$item->{line1} .= ' [E]';
+	}
```
michaelherger, line 1375: This is for display purposes only, isn't it? Eg. the suffix wouldn't be part of an album name when trying to look up MAI information?

Plugin.pm
```diff
@@ -1469,6 +1477,22 @@ sub _trackItem {
...
+	if ( $prefs->get('discNumber') && $track->{album}->{media_count} > 1 && $prefs->get('trackNumber') ) {
+		$item->{name} = "Disc $track->{media_number}/$track->{track_number}: $item->{name}";
+		$item->{line1} = "Disc $track->{media_number}/$track->{track_number}: $item->{line1}";
+	} elsif ( $prefs->get('discNumber') && $track->{album}->{media_count} > 1 ) {
+		$item->{name} = "Disc $track->{media_number}: $item->{name}";
+		$item->{line1} = "Disc $track->{media_number}: $item->{line1}";
+	} elsif ( $prefs->get('trackNumber') ) {
+		$item->{name} = "$track->{track_number}: $item->{name}";
+		$item->{line1} = "$track->{track_number}: $item->{line1}";
+	}
```
michaelherger, line 1494: "Disc" should be localizable - or not used. Why not stick closer to what LMS does for the old UIs: `DISC-TRACKNUM. TITLE`? Or even re-use `Slim::Music::Info::addDiscNumberToAlbumTitle()`?

ProtocolHandler.pm
```diff
@@ -348,6 +348,18 @@ sub getMetadataFor {
...
+		$meta->{title} .= ' [E]';
+	}
+
+	if ( $prefs->get('discNumber') && $meta->{media_count} > 1 && $prefs->get('trackNumber') ) {
+		$meta->{title} = "Disc $meta->{media_number}/$meta->{track_number}: $meta->{title}";
+	} elsif ( $prefs->get('discNumber') && $meta->{media_count} > 1 ) {
+		$meta->{title} = "Disc $meta->{media_number}: $meta->{title}";
+	} elsif ( $prefs->get('trackNumber') ) {
+		$meta->{title} = "$meta->{track_number}: $meta->{title}";
+	}
```
michaelherger, line 362: Hmm... this could have a negative impact on MAI. "Disc 1/12: My Song" would definitely not be the same as "My Song" when needed to be looked up by title.

michaelherger (commented):
I'm worried about the impact of the title changes on plugins like MusicArtistInfo.

I'd also recommend to look into how LMS is dealing with disc numbers and try to stick close to that.

darrell-k:
You're right about the impact on MAI, at least when amending `$meta->{title}` as sent to the player queue. I'm not sure either the [E] flag or the disc/track number are required there anyway, so I might just remove. I think we're OK with the album/track listings, there isn't a way into MAI from there, is there?

One interesting thing I found was that MAI does work with certain strings appended, but not others. For example when retrieving lyrics with the option to add the version to the title enabled (this is The Clash box set Sound System), _London's Burning (Remastered)_ is OK, but _London's Burning (Beaconsfield Film School Session)_  is not. Is this just the online services you use being (slightly) intelligent?

mherger:
I do clean up _some_ additions. But Remastered isn't any of them. Maybe it's indeed the service I'm calling...

darrell-k:
Actually, you do ;) 

```
sub _fetchLyrics {
	my ($args, $cb, $ecb) = @_;

	# some cleanup... music services add too many appendices
	$args->{title} =~ s/[([][^)\]]*?(deluxe|edition|remaster|live|anniversary)[^)\]]*?[)\]]//ig;
	$args->{title} =~ s/ -[^-]*(deluxe|edition|remaster|live|anniversary).*//ig;

```
Do you think we could add:

`	$args->{title} =~ s/\[E\]$//; # remove Parental Warning/Explicit flag (from Qobuz plugin)
`

darrell-k:
OK, still assuming that the little change to MAI for the "Explicit" flag is possible (see above), I've redone the track number/disc number stuff:

I'm no longer adding numbers to the track titles. I've managed to implement the more complex solution of creating playlist headers in the track listing for each disc, which puts us ahead of the Qobuz native app and will make at least one user *very* happy :)

These disc playlists will not be created when classical works processing is active and works are found, that would be extremely confusing (works split across discs, etc) even if it didn't completely break things. I've put a note about this in the settings screen, with translations using Google - I hope they're not completely meaningless!

The "Show Discs" option will also add the "disc x of y" thing  to the subsidiary lines in the album listing and the player queue. 

For track numbers, I've added metadata which persuades Material to show track numbers in the player queue/now playing screens, if the relevant option is set in the Material options. No track numbers in the album listing, but no-one has actually asked for that.

michaelherger:
Oh... f...ing inconsistencies: I was looking at other parts of the code where I do a cleanup on the _album_ title, but not the lyrics. Thanks for pointing that out!

darrell-k:
Yes, I 'd have run through that code when testing to ensure that adding (disc x of y) to the album name in the player queue didn't have any bad effects on the MAI album-based lookups.

So, any further comments/suggestions on this? Or are we about ready?

Plugin.pm
michaelherger: I'll remove that in MAI.

michaelherger:
Thanks!!

PR #48, opened 2023-06-22
Display album replay gain even when it is zero.

SamInPgh:
Found by a user in beta testing.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1234,7 +1234,7 @@ sub QobuzGetTracks {
 				type => 'text'
 			};
 			
-			if ($album->{replay_gain}) {
+			if (exists $album->{replay_gain}) {
 				push @$items,{
 					name  => sprintf( "%2.2f dB", $album->{replay_gain}),
 					label => 'ALBUMREPLAYGAIN',


Plugin.pm
```diff
@@ -1234,7 +1234,7 @@ sub QobuzGetTracks {
 				type => 'text'
 			};
 			
-			if ($album->{replay_gain}) {
+			if (exists $album->{replay_gain}) {
```
michaelherger, line 1237: I think you should use `defined` rather than `exists`, as the latter `returns true if the specified element in the hash has ever been initialized, even if the corresponding value is undefined`. And you wouldn't want to render an undefined value, would you?

SamInPgh: It would never contain an undefined value as the element is only created when one or more tracks of an album contains a valid track RG value, and it is set to the lowest of these values.

michaelherger: Even if it won't be wrong here, it's still the wrong expression, isn't it? I'm going to fix this on my end, ok?

SamInPgh: Sure. No problem.

PR #50, opened 2023-06-23
remove debug code from PR#47 (sorry).

darrell-k:


michaelherger:
Oops... should have caught that 😁.

PR #52, opened 2023-06-24
Track metadata showing wrong value for sample rate. 

SamInPgh:
Please disregard the extraneous commits which do nothing. My fork got out of sync due to my closed PR. My bad....

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1234,7 +1234,7 @@ sub QobuzGetTracks {
 				type => 'text'
 			};
 			
-			if ($album->{replay_gain}) {
+			if (exists $album->{replay_gain}) {
 				push @$items,{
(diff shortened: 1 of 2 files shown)


darrell-k:
I just posted the following on the forum:

OK, whatever is happening here is happening deep within the bowels of LMS. It seems to rest on whether the call to parseHeaders is via parseDirectHeaders or not. On two different servers, with everything player and Qobuz the same, I get different behaviours. Most probably caching differences. 

Having said this, in all my testing, Sam's change already merged on Github does resolve the issue when it exists, and doesn't do any harm when it does not. So I think we should get that released ASAP and obviously monitor here for feedback.

michaelherger:
v2.15.2 is on its way to the users. Thanks!

PR #53, opened 2023-06-26
Re-add Squeezer to list of clients supporting Weblinks

SamInPgh:
See https://github.com/kaaholst/android-squeezer/issues/799.   Verified with latest version of Squeezer on 6/26/23.


michaelherger:
Thanks! Note to myself: update my plugins, too. And LMS?

PR #56, opened 2023-09-18
Change stream quality selection

darrell-k:
Changed stream selection logic:

When the user preference is hi-res 96K, we'll stream up to 96K for all players capable of 48K and up.  The server will transcode if necessary for 48K players. (previously, we'd use the 16/44.1 stream for 48K players).

When the user preference is hi-res >96K, we'll stream up to 96K for players capable of 48K and 96K (server will transcode if necessary as above). And we'll stream up to 192K for players capable of 192K. (previously, we'd use a >96K stream for 96K players, with the server transcoding).

Tested with SB3 (48K player), Touch (96K player) and UPnPBridge->WiiM Mini (192K player).

### DIFF
--- a/API.pm
+++ b/API.pm
@@ -494,12 +494,12 @@ sub getFileInfo {
 
 	if ($format =~ /fl.c/i) {
 		$preferredFormat = $prefs->get('preferredFormat');
-		if ($preferredFormat < QOBUZ_STREAMING_FLAC_HIRES || ($maxSupportedSamplerate && $maxSupportedSamplerate < 48_000)) {
-			$preferredFormat = QOBUZ_STREAMING_FLAC;
+		if ($preferredFormat < QOBUZ_STREAMING_FLAC_96000 || ($maxSupportedSamplerate && $maxSupportedSamplerate < 48_000)) {
+			$preferredFormat = QOBUZ_STREAMING_44100;
 		}
-		elsif ($preferredFormat > QOBUZ_STREAMING_FLAC_HIRES) {
+		elsif ($preferredFormat > QOBUZ_STREAMING_FLAC_96000) {
         		if ($maxSupportedSamplerate && $maxSupportedSamplerate <= 96_000) {
-				$preferredFormat = QOBUZ_STREAMING_FLAC_HIRES;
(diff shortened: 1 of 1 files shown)


API.pm
```diff
@@ -494,11 +494,13 @@ sub getFileInfo {
 
 	if ($format =~ /fl.c/i) {
 		$preferredFormat = $prefs->get('preferredFormat');
-		if ($preferredFormat < QOBUZ_STREAMING_FLAC_HIRES || ($maxSupportedSamplerate && $maxSupportedSamplerate <= 48_000)) {
+		if ($preferredFormat < QOBUZ_STREAMING_FLAC_HIRES || ($maxSupportedSamplerate && $maxSupportedSamplerate < 48_000)) {
 			$preferredFormat = QOBUZ_STREAMING_FLAC;
 		}
 		elsif ($preferredFormat > QOBUZ_STREAMING_FLAC_HIRES) {
-			$preferredFormat = QOBUZ_STREAMING_FLAC_HIRES2;
+        		if ($maxSupportedSamplerate && $maxSupportedSamplerate <= 96_000) {
```
michaelherger, line 501: Shouldn't we use `QOBUZ_STREAMING_FLAC_HIRES2` at some point still?

Please use tabs, not spaces. Click the "Show whitespace" item at the head of the "Files changed" tab if you don't see spaces vs. tasbs. Thanks!

michaelherger (commented):
Could you please clarify what happened to using `QOBUZ_STREAMING_FLAC_HIRES2` in your change? Is it really not longer required to be used here?

API.pm
darrell-k: OK, replaced the spaces (was only on one line, I think?) - I'm using the GitHub online editor for this one, hope that's not messing anything up.

`$preferredFormat` can only be `QOBUZ_STREAMING_FLAC_HIRES2` if it's > `QOBUZ_STREAMING_FLAC_HIRES`, so that line was always superfluous.

michaelherger: Oh, right!

michaelherger:
Thanks guys!

PR #57, opened 2023-09-27
Return current sample rate / depth in metadata .

SamInPgh:
getMetaDataFor() will now always return the sample rate and depth of the currently playing Qobuz track, eliminating inconsistencies in the Now Playing technical information due to downsampling, etc.

### DIFF
--- a/ProtocolHandler.pm
+++ b/ProtocolHandler.pm
@@ -302,8 +302,13 @@ sub getMetadataFor {
 
 	if ($client && $client->playingSong && $client->playingSong->track->url eq $url) {
 		$meta->{bitrate} = $client->playingSong->bitrate if $client->playingSong->bitrate;
-		$meta->{samplerate} = $client->playingSong->pluginData('samplerate') * 1000;
-		$meta->{samplesize} = $client->playingSong->pluginData('samplesize');
+		if ( my $track = $client->playingSong->currentTrack) {
+			$meta->{samplerate} = $track->samplerate if $track->samplerate;
+			$meta->{samplesize} = $track->samplesize if $track->samplesize;
+		} else {
+			$meta->{samplerate} = $client->playingSong->samplerate if $client->playingSong->samplerate;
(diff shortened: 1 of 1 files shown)


SamInPgh:
This issue has been bothering me for awhile but I'm hopeful this fix will take care of it once and for all. Please try to break it.

ProtocolHandler.pm
```diff
@@ -302,8 +302,13 @@ sub getMetadataFor {
...
 		$meta->{bitrate} = $client->playingSong->bitrate if $client->playingSong->bitrate;
-		$meta->{samplerate} = $client->playingSong->pluginData('samplerate') * 1000;
-		$meta->{samplesize} = $client->playingSong->pluginData('samplesize');
+		if ( my $track = $client->playingSong->currentTrack) {
+			$meta->{samplerate} = $track->samplerate if $track->samplerate;
+			$meta->{samplesize} = $track->samplesize if $track->samplesize;
+		} else {
+			$meta->{samplerate} = $client->playingSong->samplerate if $client->playingSong->samplerate;
+			$meta->{samplesize} = $client->playingSong->samplesize if $client->playingSong->samplesize;
+		}
```
michaelherger, line 311: When would we have a `->currentTrack`, and when not?

SamInPgh: I can't say exactly but it seems to be created and populated at the time playback begins. Before hitting 'play' for a Qobuz track in the library, the technical info from the status query shows a 'placeholder' bitrate of 750kbps and the sample rate / size of the highest quality Qobuz stream available, as stored in the library. Shortly after starting playback, the average bitrate is calculated using the track's size and duration from the stream info and the sample rate / size values are populated in currentTrack. The values for these attributes in playingSong, however, are not updated. I didn't try to follow the logic down into LMS. The change is admittedly based on empirical observation.

michaelherger: I'll trust you'll be tweaking further in case this didn't work 😉. Thanks a lot!

PR #58, opened 2023-10-01
Fix problem jumping forward and back within a playing song.

SamInPgh:
This has been bothering me for a long time. It took awhile to figure out what was happening but, as you can see, the fix was very simple. The problem was that the bitrate would get corrupted every time a jump forward or backward was done, which caused any subsequent jumps to end up in the wrong place and further corrupt the bitrate, since the target of each jump is dependent on the bitrate. In other words, it was a mess. In my testing, I was able to jump forward and backward accurately within a song and even jump to a few seconds before the end of a gapless track while preserving the "gaplessness".


michaelherger:
Thanks! Looking good and simple. I like the that.

PR #59, opened 2023-10-09
Only write disc number & disc total for releases with > 1 disc

darrell-k:
Brings behaviour into line with local albums


darrell-k:
I was wondering - is there a way of triggering a full library update during installation of this?

michaelherger:
Thanks for this patch!

> I was wondering - is there a way of triggering a full library update during installation of this?

No automatic way to do so. And I'm usually hesitant to trigger new scans anyway.

What you could do is set a pref: if the pref doesn't exist, scan, set the pref. If it does exist, you're good. But you'd probably have to run a full wipe/rescan, which can be quite annoying if triggered automatically.

PR #61, opened 2023-11-19
French translation update

Franck-Berry:
I tried to do the best I could, but it's not easy because I don't use this plugin.
Among other things, I have doubts regarding the concepts "Untertitel", "Werke-Wiedergabelisten" and "Werktiteln". (Existing translations into other languages don't always help me.)

### DIFF
--- a/strings.txt
+++ b/strings.txt
@@ -165,13 +165,13 @@ PLUGIN_ON_QOBUZ
 PLUGIN_QOBUZ_NOT_RELEASED
 	DE	Tut uns leid: dieses Werk steht noch nicht zum Hören bereit. Es wird bald verfügbar sein.
 	EN	We're sorry, this work is not available for streaming just yet. It soon will be.
-	FR	Nous sommes désolés : cette œuvre n'est pas encore disponible à l'écoute. Elle le sera bientôt.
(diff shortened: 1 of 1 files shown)


darrell-k:
Thanks for this. I think "titres secondaires" works well.

I'm not sure about the correct French for a musical work (en) / werke (de) - is Beethoven's 5th Symphony un travail or une œuvre?

michaelherger:
I'd vote for "œuvre". But then neither my French nor knowledge of Classical music is great 😂.

Franck-Berry:
Thank you for the clarification.

I'm going to keep "titres secondaires". It's not great, but I can't think of anything better. (Maybe "titres complémentaires" ?)

I need to modify "travail". The translation I proposed is not the correct one. I will follow Michael's proposal and retain the "œuvre" target. Unlike Michael, I have some knowledge of classical music 😁 (I play a lot as an amateur musician) but I admit that without knowing the Qobuz interface and the plugin interface, it's a little hard to be sure of the context of use of each word.

I'm going to try to change that this evening.

Franck-Berry:
It's corrected.
Thank you again for your help.

michaelherger:
Thanks a lot, @Franck-Berry ! What a dedication to supporting French everywhere 😃.

PR #63, opened 2023-12-14
Improvements to replay gain.

SamInPgh:
Import album replay gain and peak gain values for possible future use and to make album gain visible in track info.

Use "album_with_tracks" cache when calculating album gain for tracks on favorited albums that have expired from the "albumInfo" cache. Formerly these albums would be played with track gain, which is incorrect behavior. I noticed this happening on albums played from the library that had been originally favorited more than 30 days ago, which is the default expiry for "albumInfo" and "trackInfo" items, whereas the "album_with_tracks" items created by the import process have a 90-day expiry AND are re-created each time a scan is performed, minimizing, if not eliminating, the chances of the problem occurring.


michaelherger:
Thanks!

PR #64, opened 2024-01-05
Include direct link to booklet in main album listing, where possible …

darrell-k:
…- takes advantage of LMS commit https://github.com/Logitech/slimserver/commit/75674a4314c71647376f75d35d21ef33d219f810

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1695,8 +1695,34 @@ sub trackInfoMenuBooklet {
 	eval {
 		my $goodies = $remoteMeta->{goodies};
 		if ($goodies && ref $goodies && scalar @$goodies) {
+		
+			# default skin or Material
+			if ( _isBrowser($client) ) {
+				if (scalar @$goodies == 1 && @$goodies[0]->{name} eq "Livret Num\xE9rique") {
+					$item = {
+						name => _localizeGoodies($client, @$goodies[0]->{name}),
+						weblink => @$goodies[0]->{url},
+					};
+				} else {
+					my $items = [];
+					foreach ( @$goodies ) {
+						if ($_->{url} =~ $GOODIE_URL_PARSER_RE) {
+							push @$items, {
+								name => _localizeGoodies($client, $_->{name}),
+								weblink => $_->{url},
+							};
+						}
+					}
+					if (scalar @$items) {
+						$item = {
+							name => cstring($client, 'PLUGIN_QOBUZ_GOODIES'),
+							items => $items
+						};
+					}
+				}
+			
 			# jive clients like iPeng etc. can display web content, but need special handling...
-			if ( _canWeblink($client) )  {
+			} elsif ( _canWeblink($client) )  {
 				$item = {
 					name => cstring($client, 'PLUGIN_QOBUZ_GOODIES'),
 					itemActions => {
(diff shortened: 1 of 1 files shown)


darrell-k:
Thinking about it, this needs to check LMS version otherwise it will break booklet viewing in Material for older LMSs

michaelherger:
I think whatever you're trying to do with `_isBrowser()` should work the same if `_canWeblink()`. The difference indeed is that 8.4 can render weblink directly, where 8.3 and older can't. So the distinction should actually be 8.4 or not 8.4 for `_canWeblink()` clients, but not the `_isBrowser()` check. Maybe all we'd need is this small change?

```diff
diff --git a/Plugin.pm b/Plugin.pm
index 0432dc2..a6c06d1 100644
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1696,7 +1672,7 @@ sub trackInfoMenuBooklet {
 		my $goodies = $remoteMeta->{goodies};
 		if ($goodies && ref $goodies && scalar @$goodies) {
 			# jive clients like iPeng etc. can display web content, but need special handling...
-			if ( _canWeblink($client) )  {
+			if ( Slim::Utils::Versions->compareVersions($::VERSION, '8.4.0') >= 0 && _canWeblink($client) )  {
 				$item = {
 					name => cstring($client, 'PLUGIN_QOBUZ_GOODIES'),
 					itemActions => {
```

darrell-k:
The problem is that the direct link doesn't work in Squeezer/OpenSqueeze, so I needed to distinguish between a browser client (Material or default) and those client apps.

On the version question, it's not all 8.4.0 is it? the change only went in a few days ago, so there will be 8.4.0s out there without the capability. 

So I was going to check `$::REVISION` as well, which leads to a question - what will that be once 8.4.0 is released? will `$::REVISION > nnnnnnnnn` still work?

michaelherger:
Would those clients return `_canWeblink()`? They shouldn't if they can't handle the links. 

And don't care about old 8.4 builds: users running 8.4 should be used to update to the latest or go back to 8.3.

darrell-k:
They are fine with the existing code which renders a submenu containing the link, but not with a direct link on the main album listing. 

```
                                $item = {
                                        name => cstring($client, 'PLUGIN_QOBUZ_GOODIES'),
                                        itemActions => {
                                                items => {
                                                        command  => [ 'qobuz', 'goodies' ],
                                                        fixedParams => {
                                                                goodies => to_json($goodies),
                                                        }
                                                },
                                        },
                                };
                        }
```

We did discuss this in LMS forum PMs last summer, at one point you said:

> Hm... as I tried to find good entry points I came to realize that even if we change XMLBrowser, the clients will have to update, too: they currently don't expect such external links in those menu responses, as this hasn't existed before. Looking at Material I think it _might_ just work. But that would be purely coincidental.

michaelherger:
> We did discuss this in LMS forum PMs last summer, at one point you said:
> 
> > Hm... as I tried to find good entry points I came to realize that even if we change XMLBrowser, the clients will have to update, too: they currently don't expect such external links in those menu responses, as this hasn't existed before. Looking at Material I think it _might_ just work. But that would be purely coincidental.

It's always good when people remember what I discussed before 🤦🏻. Was this on Github, via email, or in the forums? I don't remember why I came to the conclusion that this would not work...

darrell-k:
It was in a Forum PM thread between Me, you and Sam, entitled "Qobuz plug-in: identification of client ".

My testing indicates that you were correct to say what you did - your recent change to XMLBrowser enables direct weblinks to work with Material, but not with Squeezer or OpenSqueeze. Hence the logic in this PR.

PR #65, opened 2024-01-12
Fix #62 - Multi account support

### LINKED ISSUE #62: Multi user support
The Tidal-plugin supports adding multiple Tidal accounts and selecting a Tidal-account per player.
Has this been considered for the Qobuz-plugin? In a house with a lot of SqueezeBoxes and five users it is a valuable advantage.

michaelherger:
- [x] Allow adding more than one account
- [x] Browse mode (async) API
- [x] Let user select account to use per player
- [x] Scanner mode (sync) API
- [x] Double check cache use: make sure we don't conflict with other users
- [x] Double check prefs access for `token` and `userdata`
- [x] Remove some legacy slack while at it
- [x] Bump version number to 3.0
- [x] Require LMS 8.3 in `install.xml`


darrell-k:
Hi Michael. `canLossless` in `API/Common` is failing to detect a flac-capable account, so  I can only play MP3s. The thing is, all  Qobuz accounts are studio quality (up to 24/192) these days, so we could probably get rid of this check. Assuming  that's the case in all territories, of course.

michaelherger:
Are you saying there's no more lossy subscription for Qobuz available? That would indeed simplify things. Thanks for the feedback!

darrell-k:
That is the case in the UK, at least. You can still chose to stream MP3, but there is no account that is limited to lossy. Also,the subsciption level limited to "CD quality" has gone. Everyone can have 24/192 if they want it!

darrell-k:
I've had a trawl around various country sites (France, Chile, Argentina, US) and it seems the only subscription plan worldwide is hi-res. Seems to have been that way for 4 years!

michaelherger:
> I've had a trawl around various country sites (France, Chile, Argentina, US) and it seems the only subscription plan worldwide is hi-res. Seems to have been that way for 4 years!

Thanks for the investigation!

michaelherger:
@SamInPgh @darrell-k - I think I'd be ready of a .0 release. What do you think?

SamInPgh:
Sounds good to me. Great job on the multiple account support, by the way. I think there will be some happy users out there.

darrell-k:
Have you forgotten about my  booklet link PR?

PR #66, opened 2024-01-16
Fix cache fetch errors in getFormatForUrl()

SamInPgh:
I'm not sure if this fix might be masking another issue so please review. Without this change, I cannot get v3.0.0 working at all.

### DIFF
--- a/ProtocolHandler.pm
+++ b/ProtocolHandler.pm
@@ -95,7 +95,21 @@ sub getFormatForURL {
 		return $type;
 	}
 
-	my $info = Plugins::Qobuz::Plugin::getAPIHandler($client)->getCachedFileInfo($id || $url);
+	$id ||= $url;
+	my $api = Plugins::Qobuz::Plugin::getAPIHandler($client);
+	my $info;
+
+	# grab metadata from backend if needed, otherwise use cached values
+	if ($id && $client && $client->master->pluginData('fetchingMeta')) {
+		Slim::Control::Request::notifyFromArray( $client, [ 'newmetadata' ] ) if $client;
+		$info = $api->getCachedFileInfo($id);
+	}
+	elsif ($id) {
+		$client->master->pluginData( fetchingMeta => 1 ) if $client;
+		$info = $api->getTrackInfo(sub {
+			$client->master->pluginData( fetchingMeta => 0 ) if $client;
+		}, $id);
+	}
 
 	return $info->{mime_type} =~ /flac/ ? 'flc' : 'mp3' if $info && $info->{mime_type};
 


michaelherger:
I believe the issue is with the cache handling. I can't reproduce it, playback is working fine in my testing... Now I'm not sure whether your change _fixes_ the issue or "only" works around it. I'll try to get a better understanding, but will merge if I can't find what I'm looking for in an hour.

Thank you very much for covering me! This should teach me _not_ to rush a release before going to sleep.

michaelherger:
Even the 30s snippets are now FLAC!

SamInPgh:
Try testing with a Chromecast or Airplay device using Philippe's bridge plugins. It would probably also happen with the UPnP bridge, but I didn't try that in my testing. I think that the problem might be with the writing of the cache and it just shows up when trying to read it. But why only with the bridge players? Permissions? I agree that, unless the "real" problem, if any, can be quickly identified and addressed, we should probably push out the PR as an interim fix. It's my turn to sleep now. Good night.

michaelherger:
Maybe we could simply change the fallback result from mp3 to FLAC and that would be it. Or always just return FLAC, no other code than that!

If you are still able to reproduce the issue, please try the following:

```diff
diff --git a/ProtocolHandler.pm b/ProtocolHandler.pm
index f14ebf0..d93ede4 100644
--- a/ProtocolHandler.pm
+++ b/ProtocolHandler.pm
@@ -85,23 +85,7 @@ sub scanUrl {
        $args->{cb}->( $args->{song}->currentTrack() );
 }
 
-sub getFormatForURL {
-       my ($class, $client, $url) = @_;
-
-       my ($id, $type) = $class->crackUrl($client, $url);
-
-       if ($type =~ /^(flac|mp3)$/) {
-               $type =~ s/flac/flc/;
-               return $type;
-       }
-
-       my $info = Plugins::Qobuz::Plugin::getAPIHandler($client)->getCachedFileInfo($id || $url);
-
-       return $info->{mime_type} =~ /flac/ ? 'flc' : 'mp3' if $info && $info->{mime_type};
-
-       # fall back to whatever the user can play
-       return Plugins::Qobuz::API::Common->getStreamingFormat($client);
-}
+sub getFormatForURL { 'flc' }
```

michaelherger:
The above would be a bit too simplistic. It doesn't cover the user preferring a different streaming format. What about https://github.com/pierrepoulpe/SqueezeboxQobuz/compare/master...fix-streaming? The important part is in `Plugins::Qobuz::API::Common->getStreamingFormat()` which would now assume lossless unless any other hint was given. I think this currently might fall back to mp3 instead.

PR #67, opened 2024-01-23
Add album listing option to view detailed artist metadata.

darrell-k:
Shows consolidated artist information under a new "All  Artists" album listing option. The resulting list will show which tracks each artist had contributed to.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1410,7 +1410,7 @@ sub QobuzGetTracks {
 					}
 				}
 				my $item = {
-					name => cstring($client, 'ALL_ARTISTS'),
+					name => cstring($client, 'PLUGIN_QOBUZ_PERFORMERS'),
 					items => \@uniquePerformers,
 				};
 				push @$items, $item;
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1405,7 +1405,7 @@ sub QobuzGetTracks {
 				foreach my $item (@uniquePerformers) {
 					my @tracks = @{$tracks->{$item->{'name'}}->{'tracks'}};
 					if ( @tracks && scalar @tracks < $album->{tracks_count} ) {
-						$item->{'name'} .= scalar @tracks == 1 ? " (track " : " (tracks ";
+						$item->{'name'} .= scalar @tracks == 1 ? " (" . cstring($client, 'PLUGIN_QOBUZ_TRACK_LC') . " " : " (" . cstring($client, 'PLUGIN_QOBUZ_TRACKS_LC') . " ";
 						$item->{'name'} .= join(", ", @tracks) . ")";
 					}
 				}
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1167,8 +1167,18 @@ sub QobuzGetTracks {
 		my $workComposer;
 		my $lastDisc;
 		my $discs = {};
+		my @performers;
 
 		foreach my $track (@{$album->{tracks}->{items}}) {
+
+			if (my $trackPerformers = trackInfoMenuPerformers($client, undef, undef, $track)) {
+				my $performerItems = $trackPerformers->{items};
+				foreach my $item (@$performerItems) {
+					$item->{'track'} = $track->{'track_number'};
+				}
+				push @performers, @$performerItems;
+			}
+
 			$totalDuration += $track->{duration};
 			my $formattedTrack = _trackItem($client, $track);
 			my $work = delete $formattedTrack->{work};
@@ -1378,6 +1388,34 @@ sub QobuzGetTracks {
 				nextWindow => 'parent'
 			};
 
+			if (my $item = trackInfoMenuBooklet($client, undef, undef, $album)) {
+				push @$items, $item;
+			}
+
+			# Build a consolidated list of all artists on the album
+			my @uniquePerformers;
+			if ( scalar @performers ) {
+				my %seen = ();
+				my $tracks;
+				foreach my $item (@performers) {
+					push @{$tracks->{$item->{'name'}}->{'tracks'}}, $item->{'track'};
+					delete $item->{'track'};
+					push(@uniquePerformers, $item) unless $seen{$item->{'name'}}++;
+				}
+				foreach my $item (@uniquePerformers) {
+					my @tracks = @{$tracks->{$item->{'name'}}->{'tracks'}};
+					if ( @tracks && scalar @tracks < $album->{tracks_count} ) {
+						$item->{'name'} .= scalar @tracks == 1 ? " (track " : " (tracks ";
+						$item->{'name'} .= join(", ", @tracks) . ")";
+					}
+				}
+				my $item = {
+					name => cstring($client, 'ALL_ARTISTS'),
+					items => \@uniquePerformers,
+				};
+				push @$items, $item;
+			}
+
 			push @$items,{
 				name  => $album->{genre},
 				label => 'GENRE',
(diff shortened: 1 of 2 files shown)


Plugin.pm
```diff
@@ -1378,6 +1388,34 @@ sub QobuzGetTracks {
...
+				my $tracks;
+				foreach my $item (@performers) {
+					push @{$tracks->{$item->{'name'}}->{'tracks'}}, $item->{'track'};
+					delete $item->{'track'};
+					push(@uniquePerformers, $item) unless $seen{$item->{'name'}}++;
+				}
+				foreach my $item (@uniquePerformers) {
+					my @tracks = @{$tracks->{$item->{'name'}}->{'tracks'}};
+					if ( @tracks && scalar @tracks < $album->{tracks_count} ) {
+						$item->{'name'} .= scalar @tracks == 1 ? " (track " : " (tracks ";
```
michaelherger, line 1408: Shouldn't "track"/"tracks" be localized?

Plugin.pm
```diff
@@ -1410,10 +1448,6 @@ sub QobuzGetTracks {
 				};
 			};
 
-			if (my $item = trackInfoMenuBooklet($client, undef, undef, $album)) {
-				push @$items, $item;
-			}
```
michaelherger, line 1415: What happened to this?

michaelherger (commented):
The title says "add album menu option" - but that's not optional, is it?

Wouldn't the All Artists be added to each track rather than than to the album?

Plugin.pm
darrell-k: Yes.

Plugin.pm
darrell-k: I moved it up a bit.

darrell-k:
The option is that it is optional to click it :)

And probably better titled "add to album listing"

You can already see the individual track contributors via the "more" option from the playlist. This is leveraging the existing routine trackInfoMenuPerformers to provide an overview at the album level. I think it's particularly useful when Qobuz don't provide an album booklet, which is a lot of the time.

darrell-k:
Annoyingly, there dosn't seem to be a string for "TRACKS" (plural) in either LMS or Qobuz strings.

michaelherger:
> Annoyingly, there dosn't seem to be a string for "TRACKS" (plural) in either LMS or Qobuz strings.

There's `LCSONGS` (which stands for "lower case songs"). You could `ucfirst(string('LCSONGS'))` 😖.

darrell-k:
I actually want lower case. And I don't like "songs" - makes no sense for most classical music (though there are classical songs, a small subset of the genre).

I've created a couple of new strings in the plugin.

Plugin.pm
```diff
@@ -1378,6 +1388,34 @@ sub QobuzGetTracks {
...
+				}
+				foreach my $item (@uniquePerformers) {
+					my @tracks = @{$tracks->{$item->{'name'}}->{'tracks'}};
+					if ( @tracks && scalar @tracks < $album->{tracks_count} ) {
+						$item->{'name'} .= scalar @tracks == 1 ? " (" . cstring($client, 'PLUGIN_QOBUZ_TRACK_LC') . " " : " (" . cstring($client, 'PLUGIN_QOBUZ_TRACKS_LC') . " ";
+						$item->{'name'} .= join(", ", @tracks) . ")";
+					}
+				}
+				my $item = {
+					name => cstring($client, 'ALL_ARTISTS'),
```
michaelherger, line 1413: Ok, gave that a try to understand the experience: this is basically a "Credits" menu, isn't it? The "All Artists" string doesn't really work in German, as it's too specific and would be wrong for the engineers, mixers, and what not.

I'm using the following in MAI:

```
	EN	Credits
	FR	Crédits
	NL	Credits
```

strings.txt
```diff
@@ -536,3 +536,17 @@ PLUGIN_QOBUZ_SHOW_DISCS_DESC
 	NL	Heeft geen effect wanneer klassieke verbeteringen (zie hieronder) in gebruik zijn en er werken zijn geïdentificeerd
 	ES	No tiene efecto cuando se utilizan mejoras clásicas (ver más abajo) y se han identificado obras clásicas
 
+PLUGIN_QOBUZ_TRACKS_LC
+	DE	titel
```
michaelherger, line 540: I'm sorry, German would always be upper case 😂.

michaelherger (changes requested):
Could you please tweak the two strings? German is complicated... Other than that (and potentially other languages I don't speak) this should be good.

michaelherger:
Hmm... Green Day would only be listed as "producer" here?

Bildschirmfoto 2024-01-24 um 06 23 24

darrell-k:
Yes, they seem to be using it as an additional credits list, beyond the obvious artist. I think the labels probably provide (inconsistently) the metadata, and the streaming services don't think about it much, or at all.

By the way, I like your idea of using "Credits" as the name of the list, sounds much better than "Personnel" which is used for the playlist -- more menu entry. That sounds far too business/corporate.

So I might change it here: ths is the string definition. Interestingly, look at the DE/FR/NL entries! So "Credits" is the same in EN/FR/NL give or take an accented character? What about the German?

```
PLUGIN_QOBUZ_PERFORMERS
	DE	Performer
	EN	Personnel
	FR	Interprète
	NL	Performer
```

strings.txt
michaelherger: This still needs to be uppercase.

strings.txt
```diff
@@ -536,3 +536,17 @@ PLUGIN_QOBUZ_SHOW_DISCS_DESC
...
+	EN	tracks
+	FR	pistes
+	NL	nummers
+	ES	pistas
+
+PLUGIN_QOBUZ_TRACK_LC
+	DE	titel
```
michaelherger, line 547: Please uppercase the German Titel

darrell-k:
Thanks. Before I push again, I'm looking into the feasibility of adding the "more" option to the tracks in a Qobuz album listing (like is available for local files) so that the user does not need to add to their player queue in order to browse the track metadata. I'm guessing this starts with adding an `%actions` structure to the Qobuz track items, but any hints would be welcome.

michaelherger:
> Thanks. Before I push again, I'm looking into the feasibility of adding the "more" option to the tracks in a Qobuz album listing (like is available for local files) so that the user does not need to add to their player queue in order to browse the track metadata. I'm guessing this starts with adding an `%actions` structure to the Qobuz track items, but any hints would be welcome.

TBH: I don't know by heart... those additional menus have always been a bit odd for music services. I'd have to dig code myself, too.

SamInPgh:
> I'm looking into the feasibility of adding the "more" option to the tracks in a Qobuz album listing (like is available for local files) so that the user does not need to add to their player queue in order to browse the track metadata. 

This would be a great addition!

darrell-k:
> You'll probably have to explore XMLBrowser to figure out what it's expecting.

This is going to take some time. So I'm pushing this as it is - it stands on its own as an enhancement.

michaelherger:
Thanks!

strings.txt
```diff
@@ -285,10 +285,11 @@ PLUGIN_QOBUZ_BOOKLET
...
 PLUGIN_QOBUZ_PERFORMERS
-	DE	Performer
-	EN	Personnel
-	FR	Interprète
-	NL	Performer
+	DE	Credits
+	EN	Credits
+	FR	Crédits
+	NL	Credits
+	ES	Créditos
```
michaelherger, line 292: BTW: you know that EN, FR, ES would have been good enough? If DE wasn't found it would use EN, which is identical in this case.

PR #70, opened 2024-02-02
Info enhancements

darrell-k:
Summary of Changes:

TrackInfo: Qobuz Booklet and Qobuz Performers (ie Credits) moved within "On Qobuz" to avoid a problem when the Qobuz Cache for the track/album had expired - it is refreshed when clicking into the Track/Album info menu so will almost certainly be there before the user manages to click "On Qobuz". And after all those things are On Qobuz.

AlbumInfo: greatly enhanced to show relevant Qobuz metadata & "Manage Favourites".

These, and the matching items in the Qobuz Album view, now have search links to the record label and the personnel within the "Credits" item. Search for the recording engineer or producer! (Subject to Qobuz metadata, of course...)

The basic searches available at the top level of "On Qobuz" have been extended to include composers and works.

Once this PR is accepted, I'll write a guide to the new functionality for the forum.

### DIFF
--- a/API.pm
+++ b/API.pm
@@ -218,6 +218,22 @@ sub getArtist {
 	});
 }
 
+sub getLabel {
+	my ($self, $cb, $labelId) = @_;
+
+	$self->_get('label/get', sub {
+		my $results = shift;
+
+		$results->{albums}->{items} = _precacheAlbum($results->{albums}->{items}) if $results->{albums};
+
+		$cb->($results) if $cb;
+	}, {
+		label_id => $labelId,
+		extra     => 'albums',
+		limit     => QOBUZ_DEFAULT_LIMIT,
+	});
+}
+
 sub getArtistPicture {
 	my ($self, $artistId) = @_;
 
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -47,7 +47,7 @@ sub initGenreMap {
 }
 
 sub getCache {
-	return $cache ||= Slim::Utils::Cache->new('qobuz', 3);
+	return $cache ||= Slim::Utils::Cache->new('qobuz', 4);
 }
 
 sub getAccountList {
@@ -186,6 +186,8 @@ sub _precacheAlbum {
 			media_count => $album->{media_count},
 			duration => 0,
 			release_type => $album->{release_type},
+			label => $album->{label}->{name},
+			labelId => $album->{label}->{id},
 		};
 
 		_precacheTracks([ map {
@@ -228,7 +230,6 @@ sub _precacheTracks {
 
 sub precacheTrack {
 	my ($class, $track) = @_;
-
 	if ( !$track && ref $class eq 'HASH' ) {
 		$track = $class;
 		$class = __PACKAGE__;
@@ -259,6 +260,8 @@ sub precacheTrack {
 		track_number => $track->{track_number},
 		media_number => $track->{media_number},
 		media_count => $album->{media_count},
+		label => ref $album->{label} ? $album->{label}->{name} : $album->{label},
+		labelId => ref $album->{label} ? $album->{label}->{id} : $album->{labelId},
 	};
 
 	if ($track->{audio_info}) {
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -230,6 +230,7 @@ sub _precacheTracks {
 
 sub precacheTrack {
 	my ($class, $track) = @_;
+
 	if ( !$track && ref $class eq 'HASH' ) {
 		$track = $class;
 		$class = __PACKAGE__;
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -77,6 +77,8 @@ my $log = Slim::Utils::Log->addLogCategory( {
 use constant PLUGIN_TAG => 'qobuz';
 use constant CAN_IMAGEPROXY => (Slim::Utils::Versions->compareVersions($::VERSION, '7.8.0') >= 0);
 
+my $cache = Plugins::Qobuz::API::Common->getCache();
+
 sub initPlugin {
 	my $class = shift;
 
@@ -1513,17 +1515,20 @@ sub _albumPerformers {
 				my $sep = "-";
 				my $o;
 				for my $i ( 0 .. $#tracks ) {
-					if ( looks_like_number($tracks[$i]) && $tracks[$i+1] == $tracks[$i]+1) {
-						$o .= "$tracks[$i]$sep" if $sep;
+					my $currentValue = $tracks[$i];
+					my $currentIsNumber = looks_like_number($currentValue);
+					my $nextValue = $tracks[$i+1];
+					my $nextIsNumber = looks_like_number($nextValue);
+					if ( $currentIsNumber && $nextIsNumber && $nextValue == $currentValue+1 ) {
+						$o .= "$currentValue$sep" if $sep;
 						$sep = undef;
 					} else {
+						$o .= "$currentValue";
 						$sep = "-";
-						if ( looks_like_number($tracks[$i]) && looks_like_number($tracks[$i+1]) ) {
-							$o .= "$tracks[$i], ";
-						} elsif ( looks_like_number($tracks[$i]) && $tracks[$i+1] && !looks_like_number($tracks[$i+1]) ) {
-							$o .= "$tracks[$i]; ";
-						} else {
-							$o .= "$tracks[$i]";
+						if ( $currentIsNumber && $nextIsNumber ) {
+							$o .= ", ";
+						} elsif ( $currentIsNumber && $nextValue && !$nextIsNumber ) {
+							$o .= "; ";
 						}
 					}
 				}
@@ -1818,12 +1823,13 @@ sub albumInfoMenu {
 	my $labelId;
 	my $composers;
 	my $works;
-	my $qobuzAlbum;
+#	my $qobuzAlbum;
 	my $items = [];
 
 	if ( !%$remoteMeta && $url =~ /^qobuz:/ ) {
 		my $albumId = (split /:/, $url)[-1];
 
+		my $qobuzAlbum = $cache->get('album_with_tracks_' . $albumId);
 		getAPIHandler($client)->getAlbum(sub {
 			$qobuzAlbum = shift;
 
@@ -1831,7 +1837,10 @@ sub albumInfoMenu {
 				$log->error("Get album ($albumId) failed");
 				return;
 			}
-		}, $albumId);
+			elsif ( $qobuzAlbum->{release_date_stream} && $qobuzAlbum->{release_date_stream} lt Slim::Utils::DateTime::shortDateF(time, "%Y-%m-%d") ) {
+				$cache->set('album_with_tracks_' . $albumId, $qobuzAlbum, QOBUZ_DEFAULT_EXPIRY);
+			}
+		}, $albumId) unless $qobuzAlbum;
 
 		if ( $qobuzAlbum ) {
 			my %seen;
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1410,7 +1410,7 @@ sub QobuzGetTracks {
 					}
 				}
 				my $item = {
-					name => cstring($client, 'ALL_ARTISTS'),
+					name => cstring($client, 'PLUGIN_QOBUZ_PERFORMERS'),
 					items => \@uniquePerformers,
 				};
 				push @$items, $item;
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1823,7 +1823,6 @@ sub albumInfoMenu {
 	my $labelId;
 	my $composers;
 	my $works;
-#	my $qobuzAlbum;
 	my $items = [];
 
 	if ( !%$remoteMeta && $url =~ /^qobuz:/ ) {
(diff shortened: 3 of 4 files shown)


API.pm
```diff
@@ -218,6 +218,22 @@ sub getArtist {
...
+	$self->_get('label/get', sub {
+		my $results = shift;
+
+		$results->{albums}->{items} = _precacheAlbum($results->{albums}->{items}) if $results->{albums};
+
+		$cb->($results) if $cb;
+	}, {
+		label_id => $labelId,
+		extra     => 'albums',
+		limit     => QOBUZ_DEFAULT_LIMIT,
```
michaelherger, line 233: Interesting how I look more critically at others' code than I do at my own... I wanted to ask for validation of `$labelId`, just to figure out that I'd never validate any input in these calls either...

API/Common.pm
```diff
@@ -47,7 +47,7 @@ sub initGenreMap {
 }
 
 sub getCache {
-	return $cache ||= Slim::Utils::Cache->new('qobuz', 3);
```
michaelherger, line 50: What a pity this is required...

Plugin.pm
```diff
@@ -1390,29 +1399,8 @@ sub QobuzGetTracks {
...
-					}
-				}
-				my $item = {
-					name => cstring($client, 'PLUGIN_QOBUZ_PERFORMERS'),
-					items => \@uniquePerformers,
-				};
-				push @$items, $item;
-			}
+			# Add a consolidated list of all artists on the album
+			$items = _albumPerformers($client, $performers, $album->{tracks_count}, $items);
```
michaelherger, line 1403: This would override/remove any previously added items, as you're replacing all of `$items` instead of adding to it.

Plugin.pm
```diff
@@ -1729,11 +1751,20 @@ sub trackInfoMenu {
 				name => cstring($client, 'PLUGIN_QOBUZ_MANAGE_FAVORITES'),
 				url  => \&QobuzManageFavorites,
 				passthrough => [$args],
-			}
+			} if $args
```
michaelherger, line 1754: `$args` will always be truthy here, as we define `my $args = ()`. It's never undefined. But this should actually be assigned `{}` rather than `()`. And then the test should be `if keys %args`.

michaelherger (changes requested):
Could you please double check the commented nitpicks? Thanks!

API.pm
darrell-k: So leave as it is? :)

Plugin.pm
darrell-k: Done.

darrell-k:
OK, Addressed your comments, and now pushing, but it includes a re-done consolidated album credits listing so that it handles disc numbers for multi-disc albums, and also collapses contiguous tracks so 1,2,3,4 becomes 1-4, for example.

So I await further comments (which I really value, by the way - this is partly a Perl learning exercise for me)

API.pm
michaelherger: Yeah, just leave it as it is.

Plugin.pm
```diff
@@ -1491,6 +1484,65 @@ sub QobuzGetTracks {
...
+					} else {
+						$sep = "-";
+						if ( looks_like_number($tracks[$i]) && looks_like_number($tracks[$i+1]) ) {
+							$o .= "$tracks[$i], ";
+						} elsif ( looks_like_number($tracks[$i]) && $tracks[$i+1] && !looks_like_number($tracks[$i+1]) ) {
+							$o .= "$tracks[$i]; ";
+						} else {
+							$o .= "$tracks[$i]";
+						}
+					}
```
michaelherger, line 1528: For readability and to save at least two CPU cycles I'd store `$tracks[$I]` and `$tracks[$I+1]` in variables, as you're using these two values all over the place. Something like `$currentValue` and `$nextValue`. You could even do the same with all the `looks_like_number` evaluations. Should give a more readable block of code. Because right now I'm having a hard time figuring out all the dollar signs and increments and what not.

michaelherger: One more thing: you could remove all the checks for `looks_like_number($tracks[$I])` and replace them with a very first check against this and return early if it failed.

Plugin.pm
michaelherger: Got it, thanks!

Plugin.pm
```diff
@@ -1749,30 +1814,114 @@ sub albumInfoMenu {
...
+		getAPIHandler($client)->getAlbum(sub {
+			$qobuzAlbum = shift;
+
+			if (!$qobuzAlbum) {
+				$log->error("Get album ($albumId) failed");
+				return;
+			}
+		}, $albumId);
+
+		if ( $qobuzAlbum ) {
```
michaelherger, line 1836: Wouldn't that almost always fail, unless the album happened to be in the cache? This conditional block should be in the callback really, shouldn't it? If we wanted to check for cached values only, we should do so directly, not rely on an async call to return in a timely fashion by chance.

darrell-k: The thing is, Slim::Menu::Albuminfo/Trackinfo triggers this (or trackinfoMenu) when building the menu containing the "On Qobuz" entry, and then again when clicking on "On Qobuz". This was existing behaviour, I didn't change this.

That's why I moved everything that relies on Qobuz metadata inside the "On Qobuz" option: by the time of that second click, it will be cached, even if it had expired before.

michaelherger: Hmm... as I thought this might have been existing code, I double checked. But I still can't find it?

darrell-k: Not in the plugin - but look in Slim::Menu::TrackInfo.pm:

```
	# Get plugin metadata for remote tracks
	my $remoteMeta = {};
	if ( $track->remote && blessed($client) ) {
		my $handler = Slim::Player::ProtocolHandlers->handlerForURL($url);
		if ( $handler && $handler->can('getMetadataFor') ) {
			$remoteMeta = $handler->getMetadataFor( $client, $url );
		}
	}

```
There is no equivalent in AlbumInfo.pm, which is why I added the code in the plugin function.

michaelherger: I'm not talking about the caller, but about the above constructions where you do an async call to `getAlbum()`, but expect the `$qobuzAlbum` variable to be set in the callback immediately, continuing in a sync manner. If you expect the album to be in the cache already, why don't you read it from the cache directly? Relying on a callback to be called before the code continues is a bit of a lottery: it can work, but you never know.

michaelherger: Ok, I see that we only cache the raw request's response. So we can't read from the cache directly. But I still believe this is wrong. It might work most of the time, because (as you said) the data will be cached. But it's asking for trouble if ever we change that code. What I'd suggest is to modify `getAlbum()` to cache the results of `_precacheAlbum`, and return that in a sync manner if available, while still running the async lookup. That way you can be sure that once the call returns you have the data if it was available. In the current situation you'd only have it if it was cached _AND_ if the async calls were fast enough to call the callback in time.

I hope this clarifies things... seems all a bit confusing, sorry.

michaelherger: _If_ the cached data does exist, then you're fine. Otherwise the async HTTP request will be called, and your code continues with the check for `$qobuzAlbum` _before_ the HTTP response has arrived. LMS would continue to run, while the async call is waiting for the response. And only once that response is in, `$qobuzAlbum` would be populated - too late for the evaluation of its content.

This all might work now under normal circumstances. But if ever somebody decided to modify one of the methods you listed, the call might not return data from the cache, or might do some other async call, and you wouldn't get back `$qobuzAlbum`. Therefore I suggest you have `getAlbum()` return cached data in a sync fashion, and don't wait for the async response in that particular caller.

darrell-k: So I am understanding it reasonably well, I think. What you describe is why I moved the options requiring Qobuz metadata inside the "On Qobuz" item, as described in my original PR description. This way, we get two shots, and don't need the data until the second one. Of course, the user could still be ahead of the async request, but I don't see how that can be avoided while maintaining the requirement for all calls to the Qobuz backend to be async. 

If we don't have the data cached (or the cached data has expired) we have no choice but to go async, do we?

I did test this extensively after setting the cache expiry to 5 minutes and I didn't trip it up. But I'm happy to change `getAlbum()` - just to confirm, you'd like me to copy the cache retrieval attempt from `_get()`, return if successful, otherwise carry on into `_get()`?

michaelherger: What I'm saying is that this is prone to failure should the code chain leading to reading the cache change at some point. You can't rely on the callback populating `$qobuzAlbum` being called before the code continues. Understood - it happens to work now the way you expect it to. But you must not rely on a callback being called in a sync way. Just don't.

If you poke around the code you'll see that the scanner would cache the (almost) identical album information in `album_with_tracks_{albumId}`:

https://github.com/pierrepoulpe/SqueezeboxQobuz/blob/b655132c41f9e886604a646449b5c89886e6e3e7/Importer.pm#L123

Now you could use this cache key to check for the existence of cached data in a sync manner. If it didn't exist you still can do the async call to get it cached for the next call. 

```diff
diff --git a/Plugin.pm b/Plugin.pm
index 9dcb8a2..6247ab6 100644
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -77,6 +77,8 @@ my $log = Slim::Utils::Log->addLogCategory( {
 use constant PLUGIN_TAG => 'qobuz';
 use constant CAN_IMAGEPROXY => (Slim::Utils::Versions->compareVersions($::VERSION, '7.8.0') >= 0);
 
+my $cache = Plugins::Qobuz::API::Common->getCache();
+
 sub initPlugin {
 	my $class = shift;
 
@@ -1818,12 +1820,13 @@ sub albumInfoMenu {
 	my $labelId;
 	my $composers;
 	my $works;
-	my $qobuzAlbum;
 	my $items = [];
 
 	if ( !%$remoteMeta && $url =~ /^qobuz:/ ) {
 		my $albumId = (split /:/, $url)[-1];
 
+		my $qobuzAlbum = $cache->get('album_with_tracks_' . $albumId);
+
 		getAPIHandler($client)->getAlbum(sub {
 			$qobuzAlbum = shift;
 
@@ -1831,7 +1834,10 @@ sub albumInfoMenu {
 				$log->error("Get album ($albumId) failed");
 				return;
 			}
-		}, $albumId);
+			else {
+				$cache->set('album_with_tracks_' . $albumId, $qobuzAlbum, time() + 86400 * 90);
+			}
+		}, $albumId) unless $qobuzAlbum;
 
 		if ( $qobuzAlbum ) {
 			my %seen;
```

I'd then move the `->getAlbum() unless $qobuzAlbum;` in the else of the following `if($qobuzAlbum)`, but didn't do this here to keep the diff short. 

Does that make sense?

darrell-k: Yes, and it has helped me understand the rationale behind the code in getMetaDataFor():

```
	if ($id && $client && $client->master->pluginData('fetchingMeta')) {
		Slim::Control::Request::notifyFromArray( $client, [ 'newmetadata' ] );
		$meta = Plugins::Qobuz::API->getCachedFileInfo($id);
	}
	elsif ($id) {
		$client->master->pluginData( fetchingMeta => 1 ) if $client;

		$meta = Plugins::Qobuz::Plugin::getAPIHandler($client)->getTrackInfo(sub {
			$client->master->pluginData( fetchingMeta => 0 ) if $client;
		}, $id);
	}

```
 Should we also add a check for metadata retrieval in progress like this does?

darrell-k: Actually, not quite: that code will call `getTrackInfo()` every time unless metadata retrieval by itself is actually in progress (that is, its been called multiple in quick succession), so it mostly relies on the cache retrieval at the top of `getTrackInfo()`

But similar.

michaelherger: I believe the use cases aren't too similar: metadata lookup is something which can happen frequently, and even in bursts (eg. track change with many web UIs and other control clients connected). Menu navigation OTOH is mostly interactive, thus slow.

darrell-k: OK, pushing changes now. 

I amended your solution slightly to ensure we don't cache unreleased albums, because the metadata is not stable before release date (as we found out before https://github.com/pierrepoulpe/SqueezeboxQobuz/pull/38/files )

I didn't move `->getAlbum()` as you suggested, for the same reason. If we did this, we'd never even attempt to populate `$qobuzAlbum` for unreleased albums, which would be a shame because the function does work most of the time even when the album is uncached.

Plugin.pm
michaelherger: Could you please tweak this? Then we'd be good to go?

Plugin.pm
michaelherger: Thanks!

Plugin.pm
```diff
@@ -1491,6 +1486,68 @@ sub QobuzGetTracks {
...
+					} else {
+						$o .= "$currentValue";
+						$sep = "-";
+						if ( $currentIsNumber && $nextIsNumber ) {
+							$o .= ", ";
+						} elsif ( $currentIsNumber && $nextValue && !$nextIsNumber ) {
+							$o .= "; ";
+						}
+					}
+				}
```
darrell-k, line 1534: I thought I did:

michaelherger:
Thanks! Please ping me when you think a release is due.

darrell-k:
Now would be as good a time as any :)

BTW, congratulations on getting 8.4.0 out.

PR #71, opened 2024-02-03
French translation update

Franck-Berry:


Franck-Berry:
Hello Michael,

Why is this PR not merged?
No rush (I know shutting down mysqueezebox.com is creating a lot of work for you right now).
This is just in case there is a problem that I haven't seen.

michaelherger:
I believe the `_LC` values should stay lower case. That's what the `_LC` suffix suggests. You changed them to upper case. Would you have an example of where that would be more suitable? Maybe we'd need to change how we use those strings?

Franck-Berry:
No, I don't have an example (I don't use any streaming services). I just stupidly copied what was done on the German translation. 😁
I've adapted the translation accordingly (I'm using lowercase again).

PR #79, opened 2024-03-27
Allow grouping of artist releases by release type.

SamInPgh:
This is a change I have been wanting to make for a long time. It adds an option to group releases for an artist by release type, as defined by the iTunes and Spotify standards. With the option in effect, each release is classified as either an Album, EP or Single and grouped under one of those three headings. The order of the groups is not configurable.  However, the releases within each group will be sorted as specified in the Sort Artist Albums option. For comparison purposes, and to avoid confusion with Qobuz's somewhat random release typing, the Qobuz release type is now displayed in the Album Info page following the Genre, and is labeled as "Qobuz Release Type". All string translations are straight from Google Translate as I am monolingual.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -46,6 +46,7 @@ $prefs->init({
 	useClassicalEnhancements => 1,
 	parentalWarning => 0,
 	showDiscs => 0,
+	groupReleases => 0,
 });
 
 $prefs->migrate(1,
@@ -581,8 +582,10 @@ sub QobuzArtist {
 			return;
 		}
 
+		my $groupByReleaseType = $prefs->get('groupReleases');
+
 		my $items = [{
-			name  => cstring($client, 'ALBUMS'),
+			name  => $groupByReleaseType ? cstring($client, 'PLUGIN_QOBUZ_RELEASES') : cstring($client, 'ALBUMS'),
 			# placeholder URL - please see below for albums returned in the artist query
 			url   => \&QobuzSearch,
 			image => 'html/images/albums.png',
@@ -618,28 +621,59 @@ sub QobuzArtist {
 		if ($artist->{albums}) {
 			my $albums = [];
 
+			# group by release type if requested
+			if ($groupByReleaseType) {
+				for my $album ( @{$artist->{albums}->{items}} ) {
+					if ($album->{duration} >= 1800 || $album->{tracks_count} > 6) {
+						$album->{release_type} = "Album";
+					} elsif ($album->{tracks_count} < 4) {
+						$album->{release_type} = "Single";
+					} else {
+						$album->{release_type} = "Ep";
+					}
+				}
+			}
+
 			# sort by release date if requested
 			my $sortByDate = $prefs->get('sortArtistAlbums');
 
 			$artist->{albums}->{items} = [ sort {
 				if ($sortByDate) {
-					return $sortByDate == 1 ? $b->{released_at}*1 <=> $a->{released_at}*1 : $a->{released_at}*1 <=> $b->{released_at}*1;
+					return $sortByDate == 1 ? ( $a->{release_type} cmp $b->{release_type} || $b->{released_at}*1 <=> $a->{released_at}*1 )
+											: ( $a->{release_type} cmp $b->{release_type} || $a->{released_at}*1 <=> $b->{released_at}*1 );
 				}
 				else {
-					# push singles and EPs down the list
-					if ( ($a->{tracks_count} >= 4 && $b->{tracks_count} < 4) || ($a->{tracks_count} < 4 && $b->{tracks_count} >=4) ) {
-						return $b->{tracks_count} <=> $a->{tracks_count};
-					}
-
-					return lc($a->{title}) cmp lc($b->{title});
+					return $a->{release_type} cmp $b->{release_type} || lc($a->{title}) cmp lc($b->{title});
 				}
 
 			} @{$artist->{albums}->{items} || []} ];
 
+			my $lastReleaseType = "";
+
 			for my $album ( @{$artist->{albums}->{items}} ) {
 				next if $args->{artistId} && $album->{artist}->{id} != $args->{artistId};
+				if ($album->{release_type} ne $lastReleaseType) {
+					$lastReleaseType = $album->{release_type};
+					my $relType = "";
+					if ($lastReleaseType eq "Album") {
+						$relType = cstring($client, 'ALBUMS');
+					} elsif ($lastReleaseType eq "Ep") {
+						$relType = cstring($client, 'RELEASE_TYPE_EPS');
+					} elsif ($lastReleaseType eq "Single") {
+						$relType = cstring($client, 'RELEASE_TYPE_SINGLES');
+					} else {
+						$relType = "Unknown";  #should never occur
+					}
+
+					push @$albums, {
+						name => $relType,
+						image => 'html/images/albums.png',
+						type => 'text'
+					} ;
+				}
 				push @$albums, _albumItem($client, $album);
 			}
+
 			if (@$albums) {
 				$items->[0]->{items} = $albums;
 				delete $items->[0]->{url};
@@ -1386,6 +1420,10 @@ sub QobuzGetTracks {
 				name  => $album->{genre},
 				label => 'GENRE',
 				type  => 'text'
+			},{
+				name => $album->{release_type} =~ /^[a-z]+$/ ? ucfirst($album->{release_type}) : $album->{release_type},
+				label => 'PLUGIN_QOBUZ_RELEASE_TYPE',
+				type => 'text'
 			},{
 				name  => Slim::Utils::DateTime::timeFormat($album->{duration} || $totalDuration),
 				label => 'ALBUMLENGTH',
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -620,16 +620,23 @@ sub QobuzArtist {
 		# use album list if it was returned in the artist lookup
 		if ($artist->{albums}) {
 			my $albums = [];
-
+			my $numAlbums = 0;
+			my $numEps = 0;
+			my $numSingles = 0;
+			
 			# group by release type if requested
 			if ($groupByReleaseType) {
 				for my $album ( @{$artist->{albums}->{items}} ) {
+					next if $args->{artistId} && $album->{artist}->{id} != $args->{artistId};					
 					if ($album->{duration} >= 1800 || $album->{tracks_count} > 6) {
 						$album->{release_type} = "Album";
+						$numAlbums++;
 					} elsif ($album->{tracks_count} < 4) {
 						$album->{release_type} = "Single";
+						$numSingles++;
 					} else {
 						$album->{release_type} = "Ep";
+						$numEps++;
 					}
 				}
 			}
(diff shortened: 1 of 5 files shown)


SamInPgh:
After removing the album icon from the group heading, this is what the artist releases screen looks like. I'm open to cosmetic changes to the headings or even making each of them a playlist, i.e. add/play the EP's only, etc. However, I personally don't see much value in the latter. I don't know why the text in the headings is paler than the release names in the item entries. If anything, I'd like it to be bolder to make them stand out more.

Screenshot_20240326-231937

michaelherger:
Haven't looked at the code yet, just some thoughts (I'm a bit stretched thin...)

* you could probably "steal" strings from LMS itself, where we have the same feature
* I personally don't like those headers - but that's me, and I'm not using the plugin myself. So... but didn't we have headers somewhere else already? I'd make them look the same
* I tend to do similar things similarly where possible. In LMS we create sub menus for the release types. So I'd go with sub menus here, too. Just for the sake for consistency.
* Always test on all UIs: Default skin, Squeezeplay, Softsqueeze or Classic (for the old two line display), Material. The lighter tone certainly is some Material specific design decision.

darrell-k:
I believe that Material is using a lighter font for simple text items, and a heavier font for links. It might be worth seeing what Craig does for the Release Type subheadings for the local library, which appear bold and blue (though that may be part of Material's specific release type handling which replaces submenus with subheadings for the main release types, and pushes other categories into the "More" menu at the top).

With the classical enhancements, we ended up with just the text for the work headings, which isn't greatly satisfactory when used with Default Skin or Squeezer etc because LMS delivers a "blank cd" icon for those list elements.

On the preferences, we could follow the main LMS release type option for release types.

SamInPgh:
> Haven't looked at the code yet, just some thoughts (I'm a bit stretched thin...)
> 
> * you could probably "steal" strings from LMS itself, where we have the same feature

That's exactly what I'm doing, where applicable.

> * I personally don't like those headers - but that's me, and I'm not using the plugin myself. So... but didn't we have headers somewhere else already? I'd make them look the same

I agree. I didn't like them either. See below for new format.

> * I tend to do similar things similarly where possible. In LMS we create sub menus for the release types. So I'd go with sub menus here, too. Just for the sake for consistency.

I disagree there. It just adds unnecessary complexity in this use case. Additionally, I'm not really looking for complete consistency here, as the Spotify/iTunes release types being used to group the releases are (thankfully) different, and simpler, than those used by Qobuz --- Albums, EPs and Singles. No Boxed Sets, Best Ofs, and whatever other arbitrary categorizations Qobuz comes up with, not to mention the many inaccuracies in the basic three types. This feature is simply to make it easier distinguish the "real" albums from the many eps and singles that now clutter the Albums by Artist display. It is already making my life much easier.

> * Always test on all UIs: Default skin, Squeezeplay, Softsqueeze or Classic (for the old two line display), Material. The lighter tone certainly is some Material specific design decision.

Business as usual. And the lighter tone in Material is now gone due to adding the album icon back to the group line after remembering the "trick" to make it non-clickable, which is to omit the "type ==> 'text'" attribute from the item entry.

SamInPgh:
> I believe that Material is using a lighter font for simple text items, and a heavier font for links. It might be worth seeing what Craig does for the Release Type subheadings for the local library, which appear bold and blue (though that may be part of Material's specific release type handling which replaces submenus with subheadings for the main release types, and pushes other categories into the "More" menu at the top).
> 

I have changed the group header to include the number of entries in the category, analogous to what MS does. If I could figure out how to make it blue, I would do that too.  ;-)

> With the classical enhancements, we ended up with just the text for the work headings, which isn't greatly satisfactory when used with Default Skin or Squeezer etc because LMS delivers a "blank cd" icon for those list elements.
> 

Yeah. I remembered how to include an icon without making it clickable. Will upload the new look in a screenshot from my phone in a minute or two.

> On the preferences, we could follow the main LMS release type option for release types.

I'm not sure what you mean here, but if you're talking about somehow tying this to how release types are handled in the library and beyond, that is not happening. This is a separate checkbox preference item in the Qobuz plugin.

SamInPgh:
Thanks for the input, guys! Here are screenshots showing the new optional grouping of Artist Albums by release type and the menu preference to enable it, which defaults to 'off'. It looks much better, IMHO. Additional feedback welcomed. (Quick fact: did you know that there are 90 Bob Dylan albums currently on Qobuz but only 2 EPs and 3 singles?) 😉

Screenshot_20240327-142811

Screenshot_20240327-152230

SamInPgh:
Just to give a couple of examples of how useless Qobuz's release types are using the Billy Strings releases above, the one entitled "Billy Strings | OurVinyl Sessions" is clearly an EP, as it contains only 4 songs totaling 19 minutes. And, as if the title wasn't enough of a clue, the release "Billy Strings - EP", containing 6 tracks running a total of 19:41, is also an EP. However, both are tagged as "album" by Qobuz.

michaelherger:
That's looking much nicer with the album icons!

I was wondering about the settings wording: wouldn't mentioning Spotify and iTunes in the Qobuz plugin be confusing some users? I know we do so in the main LMS settings. But there we mention the other methods, too, and have a link to some source (don't we?). Maybe this could be something simpler like "Override Qobuz' release type and group into albums/EPs/singles"?

And I would agree with @darrell-k: why not use the global setting?

I still believe consistency is a good thing 🙂.

Plugin.pm
```diff
@@ -617,29 +620,70 @@ sub QobuzArtist {
...
 
+			my $lastReleaseType = "";
+
 			for my $album ( @{$artist->{albums}->{items}} ) {
 				next if $args->{artistId} && $album->{artist}->{id} != $args->{artistId};
+				if ($album->{release_type} ne $lastReleaseType) {
+					$lastReleaseType = $album->{release_type};
+					my $relType = "";
+					my $relNum = 0;
+					if ($lastReleaseType eq "Album") {
```
michaelherger, line 666: I'd do case insensitive comparisons. It's too easy to get it wrong.

strings.txt
```diff
@@ -540,6 +540,20 @@ PLUGIN_QOBUZ_SHOW_DISCS_DESC
 	NL	Heeft geen effect wanneer klassieke verbeteringen (zie hieronder) in gebruik zijn en er werken zijn geïdentificeerd
 	ES	No tiene efecto cuando se utilizan mejoras clásicas (ver más abajo) y se han identificado obras clásicas
 
+PLUGIN_QOBUZ_GROUP_RELEASES
```
michaelherger, line 543: SETUP_GROUP_BY_RELEASE_TYPES_1

strings.txt
```diff
@@ -540,6 +540,20 @@ PLUGIN_QOBUZ_SHOW_DISCS_DESC
...
+	DE	Gruppieren Sie Künstlerveröffentlichungen nach Typ
+	EN	Group artist releases by type
+	FR	Regrouper les sorties d'artistes par type
+	NL	Groepeer releases van artiesten op type
+	ES	Lanzamientos de artistas grupales por tipo
+
+PLUGIN_QOBUZ_GROUP_RELEASES_DESC
+	DE	Gruppieren Sie Künstlerveröffentlichungen nach Typ, wie von Spotify und iTunes definiert: d. h. Alben, EPs und Singles
```
michaelherger, line 551: "Nach Veröffentlichungstyp gruppieren, wie von..."

SamInPgh:
> That's looking much nicer with the album icons!
> 
> I was wondering about the settings wording: wouldn't mentioning Spotify and iTunes in the Qobuz plugin be confusing some users? I know we do so in the main LMS settings. But there we mention the other methods, too, and have a link to some source (don't we?). Maybe this could be something simpler like "Override Qobuz' release type and group into albums/EPs/singles"?
> 
> And I would agree with @darrell-k: why not use the global setting?
> 
> I still believe consistency is a good thing 🙂.

First of all, thanks. I agree that it looks much better with the icons (and the number of entries for each release type). 

Now to the rebuttal: If you really believe, as I do, that consistency is a good thing, then you shouldn't look very closely at Qobuz release type tagging, which is a completely inconsistent, arbitrary and error-filled mess. I would like to address both of your objections in detail, but I am currently ill and medicated so bear with me.

- The settings wording:  I'm okay with replacing the references to iTunes and Spotify with something like "industry standard methods". I think that providing a link to a web site is unnecessary though. All we are talking about here is a way to display an artist's releases in a more organized fashion that doesn't require the user to guess which ones are singles, ep's and albums which, in today's industry, is getting more difficult due to the rising popularity of EPs and Singles (which, if they are the "title" song of an album, are difficult, if not impossible, to distinguish from the album itself in a list of "albums"). Which brings up another feature of the change: If the user selects the "Group artist albums by release type" option, the heading for the list becomes "Releases" instead of "Albums". Without the option selected, the heading remains "Albums", with all release types mixed together, just as before.

- Using the global setting:  This is purely a plugin-specific option for displaying Qobuz content, not just a user's favorites but all artist releases. It has nothing to do with how the LMS library is to be maintained and organized. To be quite honest, I will probably turn off release type support in LMS once this is implemented as it is currently a mess with Qobuz content. The only way I would use it is if all Qobuz releases were recategorized in the "guessing" algorithm and treated as not containing a release type at all. Then and only then would there be at least a modicum of "consistency". I just wish I knew what a mess the Qobuz release typing was before all the effort was put into supporting random occurrences of "boxset", "bestof", etc. And exactly what is an "epmini" anyway, as compared to a basic  "ep"? Is there an "epmaxi"?

Sorry. The meds are kicking in....  The bottom line is that I'm flexible on the wording of the option setting but not on forcing Qobuz users to have all their library content tagged with release type data which, in the case of Qobuz titles, will often be incorrect, just so they can view artist releases in a more user-friendly manner.

strings.txt
SamInPgh: What am I missing here? Are you suggesting changing the name of the string from "PLUGIN_QOBUZ_GROUP_RELEASES" to "SETUP_GROUP_BY_RELEASE_TYPES_1"? If so, why? And shouldn't it start with "PLUGIN_QOBUZ_" for "consistency"?   ;-)
Maybe it's the meds but I'm confused...

strings.txt
SamInPgh: Okay.

Plugin.pm
SamInPgh: Are you aware that these literals are temporary values assigned to the album release types while traversing the array a few lines previous to this in preparation for the sort? I could have used 'a', 'b' and 'c' or even 1, 2 and 3 but I thought the use of the literals would be more self-documenting. I'm open to specific suggestions though.

strings.txt
SamInPgh: Done.

strings.txt
michaelherger: What I wanted to suggest is to use that string as a base, to keep the wording consistent (see eg. the German version, which differs considerably).

Plugin.pm
michaelherger: Understood! But as it's easy to get a literal string comparison wrong without noticing I suggested using constants: those would throw an error upon startup if you introduced a typo.

SamInPgh:
Here's an interesting find. There are only two Frank Zappa releases on Qobuz that the iTunes/Spotify algorithm (and the music industry as a whole) identifies as EPs, both of which Qobuz tags otherwise. The first is "A Very Zappa Birthday" (listed on Apple Music as "A Very Zappa Birthday - EP"), which Qobuz classifies as a "Compilation" even though all 6 tracks are written and performed by Mr. Zappa. The second is probably the only Zappa release classified (at least on Qobuz) as Classical, titled "Fanfares and Fuckoffs", containing 5 tracks with a total run time of under 5 minutes, which Qobuz tags as an "Album" and for which the plugin dutifully (and somewhat comically) creates Work headings. I'm sure it's an essential part of every true Igor Stravinsky fan's collection. 😂

Screenshot_20240328-232118

SamInPgh:
Okay. Let's do a reset here. I've lost track of which comments are style suggestions only, and which are issues preventing the PR from being merged. Some may very well be both. Please clarify so that I can update to the latest plugin version without losing this feature, which I find makes the plugin infinitely more usable (only a slight exaggeration). Thanks.

michaelherger:
I'm sorry for being vague... Please change that literal comparison to use constants instead (and yes, those values then can be just 1, 2, 3 😁). Then we're good to merge.

SamInPgh:
> I'm sorry for being vague... Please change that literal comparison to use constants instead (and yes, those values then can be just 1, 2, 3 😁). Then we're good to merge.

Thanks. Done.

michaelherger (approved):

michaelherger:
Now feel the new power and click that "Merge pull request" button 😉. Thanks a lot for your effort and patience!

Issue #77, opened 2024-03-20
New playlists are not imported


th2j:
Hello

new playlists in qobuz are not imported in LMS
i had 142 playlists, then i had 2 new playlists, rescaned LMS and i am still at 142
even if i make "clear database and rescan", i am still at 142 playlists
and if i delete a playlist in qobuz and rescan, it stays in LMS

Thanks
Thomas

michaelherger:
Do you import them to your library (can be turned off either in TIDAL itself or in OMLI - don't remember...).

Always check your scanner.log. Maybe there's a hint about the failure?

th2j:
i delete all the db and now it is ok
sorry fot bothering you

th2j:
When i delete the db, the rescan was ok

But i tried to add another playlist in qobuz, and LMS doesn't import it
i can't see nothing in the scanner.log

Thanks

michaelherger:
Don't you think it's a caching issue? No need to delete the db file, but just wait an hour or so? By enabling info logging for plugin.qobuz you would hopefully see whether it's using data from the cache rather than requesting it afresh.

th2j:
I rescan this morning, after all the night, and the playlists are not updated,

darrell-k:
Should be corrected with https://github.com/LMS-Community/plugin-Qobuz/pull/80

th2j:
Ok thanks

PR #80, opened 2024-03-28
Correction of bug in playlist import: see issue #77

darrell-k:
Calling `Plugins::Qobuz::API::Common->username` without a parameter was screwing up the `$params` hash passed to `_get.` The username doesn't seem to be needed, but this is how it's done in `Plugin.pm`, so just replicating.

### DIFF
--- a/API/Sync.pm
+++ b/API/Sync.pm
@@ -120,7 +120,7 @@ sub myPlaylists {
 
 	my $playlists = $class->_get('playlist/getUserPlaylists', $userId, {
 		# TODO - need to pass the userId or configuration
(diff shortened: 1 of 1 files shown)


API/Sync.pm
```diff
@@ -120,7 +120,7 @@ sub myPlaylists {
 
 	my $playlists = $class->_get('playlist/getUserPlaylists', $userId, {
 		# TODO - need to pass the userId or configuration
```
michaelherger, line 122: Why would I leave that comment but not do it?!? Oh my... Thanks for looking into this!

PR #81, opened 2024-04-05
Import Qobuz Works tag into LMS

darrell-k:
A very simple change to get us started. Will almost certainly need enhancing in order to deal with Qobuz items which should have a work tag, but don't, much like the changes we made a year ago for the Qobuz classical enhancements in respect of plugin native album display for classical music.

But no harm in merging this as a first cut.


michaelherger (approved):
Assuming you've tested this, and those values are defined every now and then, this is looking good to me.

PR #90, opened 2025-01-06
Further changes to incorporate the inclusion and display of all main artists to match what is shown in the Qobuz application.

SamInPgh:


michaelherger (commented):
I think that's fine.

But I obviously would have implemented it differently 😄. Rather than have to branches doing the same thing for either the artist, or a list of artists, I would have created a list of artists, either extracted from the artists lists if available, or with a single artist item. Then I'd have iterated through that list (short or long) to add the items. That way you'd only have to implement adding the label once. I always try to avoid doing the same thing in different implementations as this risks to produce inconsistencies when you fix one, but not the other implementation.

SamInPgh:
Thanks for the feedback. Yes, I saw that one coming even as I wrote it. 😉  Let me look at it again. It comes down to my lack of comfort with list processing, combined with my rapidly deteriorating mental faculties. 👴 (And maybe just a dash of laziness.)

SamInPgh:
Oops! I mistakenly "Squashed and merged" all the changes from this pull request while trying to promote the latest commit. I am good with the merge if there are no objections though.

michaelherger:
No worries. Git history is a bit confusing. But other than that we should be good. Or you'll fix it next time 😉.

PR #91, opened 2025-01-07
Address the condition where an empty artist list exists for an album. Found in testing.

SamInPgh:


Plugin.pm
```diff
@@ -1437,13 +1437,16 @@ sub QobuzGetTracks {
 		}
 
 		my @artistList = ();
-		if (ref $album->{artists}) {  # get all main artists
+		my $numArtists = 0;
```
michaelherger, line 1440: You don't need a new variable. Just check for the length of the `@artistList` instead? 

```perl
if (!scalar @artistList) {  # no main artists found
```

SamInPgh: Oh. Right. Doh! 🤦‍♂️

SamInPgh: Done.

michaelherger (approved):

Importer.pm
```diff
@@ -305,26 +305,45 @@ sub _ignorePlaylists {
 	return $class->can('ignorePlaylists') && $class->ignorePlaylists;
 }
 
+# Cache this preference
+my ($_splitList, $_gotSplitList);
```
michaelherger, line 309: As this is the importer, which would be run, then quit, you don't need to care about split list changes. Just read the pref here once, and use it. No need to check using `$_gotSplitList` in the scanner, as it must not change during a scan run.

Importer.pm
```diff
@@ -305,26 +305,45 @@ sub _ignorePlaylists {
...
 sub _prepareTrack {
 	my ($album, $track) = @_;
 
 	my $url = Plugins::Qobuz::API::Common->getUrl(undef, $track) || return;
 	my $ct  = Slim::Music::Info::typeFromPath($url);
 
+	if (!$_gotSplitList) {
+		$_splitList = preferences('server')->get('splitList');
+		$_gotSplitList = 1;
+	}
```
michaelherger, line 320: Not needed. See above.

Importer.pm
```diff
@@ -305,26 +305,45 @@ sub _ignorePlaylists {
...
+				$artist .= ($numArtists ? $_splitList : '') . $artists->{name};
+				$artistId .= ($numArtists ? $_splitList : '') . $artists->{id};
+				$numArtists++;
+			}
+		}
+	}
+	if (!$numArtists && ref $album->{artist}) {
+		$artist = $album->{artist}->{name};
+		$artistId = $album->{artist}->{id};
 	}
```
michaelherger, line 338: Haven't I seen this before?... not exactly, but similar enough to not be implemented twice: share the code with `Plugins::Qobuz::Plugin::QobuzGetTracks()`, and read the returned list length instead of counting, join them?

michaelherger (changes requested):
Please re-use the artist evaluation code you already created. You can then iterate over the returned artists list to get the joined id and artist string.

Importer.pm
SamInPgh: Agreed. This was part of a cut-and-paste job that I didn't think through thoroughly.

Importer.pm
SamInPgh: Done.

Importer.pm
SamInPgh: I understand your point but not sure I agree. As you point out, the the situations are similar but not exactly the same.  I was going for efficiency here. Why not just build the $artist and $artistId while iterating the first list instead of creating another list and then iterating it too, especially when this routine is performed potentially thousands of times during the scan?

michaelherger: Efficiency claims should be backed by some measurements. Until then I'd say maintenance of a single implementation certainly is more efficient than maintaining two implementations of basically the same. 😂

```perl
my @artists = Plugins::Qobuz::SomeModule::getThatArtistsList($album);
$artist = join($_splitList, map { $_->{name} } @artists);
$artistId = join($_splitList, map { $_->{id} } @artists);
```

This certainly isn't considerably slower than your code. We're talking about lists of 1-2 items on average I'd say. Performance isn't critical, but maintenance is. And three lines instead of more than a dozen are so much easier to read, aren't they?

SamInPgh: You make a convincing argument. 😉 And thanks for the mini-tutorial on list processing with Perl. I think I need to take the full class. 🤔

SamInPgh: Done. Thanks, as always, for your patience and good humor. 😉

Importer.pm
```diff
@@ -305,26 +305,28 @@ sub _ignorePlaylists {
...
 
 	$album->{release_type} = 'EP' if lc($album->{release_type} || '') eq 'epmini';
 
 	my $attributes = {
 		url          => $url,
 		TITLE        => Plugins::Qobuz::API::Common->addVersionToTitle($track),
-		ARTIST       => $artist || $album->{artist}->{name},
-		ARTIST_EXTID => 'qobuz:artist:' . ($artistId || $album->{artist}->{id}),
+		ARTIST       => $artist,
+		ARTIST_EXTID => 'qobuz:artist:' . $artistId,
```
michaelherger, line 329: Oh, `$artistId` here would potentially be a list of IDs. The map above then would have to add the `qobuz:artist:` in front of _each_ id.

```perl
$artistId = join($_splitList, map { 'qobuz:artist:' . $_->{id} } @artistList);
```

Right?

Importer.pm
michaelherger: > And thanks for the mini-tutorial on list processing with Perl. I think I need to take the full class. 🤔

Just make sure you test my suggestions, because I most likely didn't 😬.

SamInPgh: No offense, but I test ALL changes, even my own. 😆

Importer.pm
SamInPgh: I honestly don't know. I was taking a leap of faith there.  All I can tell you is that the secondary artists all seem to be fully functional and searchable in the library. I think maybe we could use some input from @darrell-k on this subject.

darrell-k: I've downloaded this PR's changes, and will check it out, probably tomorrow now.

SamInPgh: Thanks!

michaelherger: Hmm... I see that we would indeed sometimes split the `extid` value - but always by `,`. You don't use the commas a separator, do you?

But then the `map` still is valid: whatever code later would do, `qobuz:artist:123abc;xyz987;a12bc3` certainly wouldn't be recognised as three external IDs.

SamInPgh: No, I use the default ";". Maybe I'll try using a comma for the extid separator only. Where would the incorrect artist id's be likely to show up as a problem? Navigating through the library and playing multi-artist content, everything looks fine to me, but maybe I'm just not looking in the right places. The only anomaly I've seen is that Material shows all/both artists in the library view but only the first one in Now Playing and the Queue, despite having selected the "Show all artists" option in the plugin prefs.

michaelherger: The code I was referring to actually is about storing more than one external ID with a single contributor, eg. from different music services. That's not the same problem as we're discussing here.

Would the additional artists show up even if they didn't exist in any other parts of your collection? Eg. no local file, no other music service? And would they have a valid `extid` in the database?

SamInPgh: > The code I was referring to actually is about storing more than one external ID with a single contributor, eg. from different music services. That's not the same problem as we're discussing here.
> 
> Would the additional artists show up even if they didn't exist in any other parts of your collection? Eg. no local file, no other music service?

Yes, they do.

 > And would they have a valid `extid` in the database?

I don't know how to check that. Can you give me a pointer?

By the way, changing the artistid separator to ',' caused them both to show up in the JSON debug for the Material Skin queue, but only the first artist is there, e.g.

{"artist_ids":"20515, 20119","extid":"qobuz://161994387.flac","artist":"Stone Gossard"

Plugin.pm
```diff
@@ -2301,6 +2301,20 @@ sub _isMainArtist {  # determine if an artist is a main artist on the release
 	return 0;
 }
 
+sub _isMainArtistByName {  # determine if an artist is a main artist on the release
+	my ($artistName, $album) = @_;
+
+	if ($album->{artist} && lc($album->{artist}->{name}) eq $artistName) {
```
michaelherger, line 2307: is the `$artistName` always in lowercase, or should `lc()` it, too?

Importer.pm
michaelherger: Do you speak SQL? The contributors table would have an `extid` column.

Or get something like [SQLiteStudio](https://sqlitestudio.pl) for an easy, visual tool to open `library.db`.

SamInPgh: Not really. I'll wait for @darrell-k to weigh in on this. He speaks SQL quite fluently. 😉

Plugin.pm
SamInPgh: Good point. Will do.

SamInPgh: Done.

Importer.pm
michaelherger: Ok, that will need more work on the LMS side. Could you please create a ticket in https://github.com/LMS-Community/slimserver/issues/new? Thanks!

Importer.pm
SamInPgh: Done.

Importer.pm
SamInPgh: Will do.

michaelherger (approved):

Importer.pm
```diff
@@ -305,26 +305,29 @@ sub _ignorePlaylists {
...
 	my ($artist, $artistId);
-	if (ref $album->{artists}) {
-		($artist, $artistId) = map { $_->{name}, $_->{id} } grep {
-			$_->{roles} && grep(/main-artist/, @{$_->{roles}})
-		} @{$album->{artists}};
-	}
+	
+	my @artistList = Plugins::Qobuz::API::Common->getMainArtists($album);
+	$artist = join($_splitList, map { $_->{name} } @artistList);
+#	$artistId = join(',', map { $_->{id} } @artistList);
```
michaelherger, line 321: Please remove this line, as it would need to be modified anyway (to create a full ID with the the `qobuz:..` prefix).

Importer.pm
darrell-k: > I've downloaded this PR's changes, and will check it out, probably tomorrow now.

@SamInPgh do you still have questions you'd like me to look into, or have you resolved things already? Sorry, I got a bit diverted yesterday.

michaelherger: @SamInPgh reverted to having a single main user, as LMS doesn't have the code (yet) to split external IDs into multiple contributors. It's something we should try to address in 9.1.

SamInPgh: All requested changes have been made. I'd like to merge this PR if there are no objections.

Importer.pm
SamInPgh: Done.

Importer.pm
michaelherger: I approved the pull request already. Are you saying you don't have write access to merge yourself? I can certainly do it for you, but thought you had the permissions to do it.

SamInPgh: Sorry, Michael. I missed the approval. Thanks for all your help on these changes. I will merge it now.

SamInPgh: > Ok, that will need more work on the LMS side. Could you please create a ticket in https://github.com/LMS-Community/slimserver/issues/new? Thanks!

Done. But my issue is not nearly as exciting as the one before it. No "porn sounds" involved. 😂

PR #93, opened 2025-01-17
Fix to Common.pm

SamInPgh:
It turns out that, in some cases, the primary artist is not contained in the main artist array, so always include it first in the list and filter it out when iterating the array.

### DIFF
--- a/API/Common.pm
+++ b/API/Common.pm
@@ -306,6 +306,23 @@ sub addVersionToTitle {
 	return $track->{title};
 }
 
+sub getMainArtists {
+	my ($class, $album) = @_;
+	
+	my @artistList = ();
+	if (ref $album->{artists} && scalar @{$album->{artists}} ) {  # get all main artists
+		for my $artist ( @{$album->{artists}} ) {
+			if (grep(/main-artist/, @{$artist->{roles}})) {
+				push @artistList, $artist;
+			}
+		}
+	}
+	if (!scalar @artistList) {  # no main artists found, use the primary artist
+		push @artistList, $album->{artist};
+	}
+	return @artistList;
+}
+
 sub getStreamingFormat {
 	my ($class, $track) = @_;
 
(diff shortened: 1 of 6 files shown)


SamInPgh:
I have no idea why every time I push a change it results in all these extra commits, 19 in this case. I know I'm doing something wrong but don't know what.

michaelherger (commented):
Did you test this with the Aarmin van Buren issue reported in the forum. I believe in that case the `$album->{artist}` was _not_ what we were looking for, and therefore we didn't get the expected list. Unfortunately I won't have time to look into this in more details before Sunday/Monday. Just wanted to ask you to double check we don't revert that fix

michaelherger:
> I have no idea why every time I push a change it results in all these extra commits, 19 in this case. I know I'm doing something wrong but don't know what.

I guess that your `master` branch is slightly out of synch with `upstream` (this repository). I usually try to keep them in sync at all times, doing a `git reset --hard upstream/master` or similar (I'm using SourceTree to do that). And whenever I want to submit a PR, I create a new branch based on the upstream `master`. I don't (most of the time anyway) edit `master` directly and use it for the PR.

SamInPgh:
> Did you test this with the Aarmin van Buren issue reported in the forum. I believe in that case the `$album->{artist}` was _not_ what we were looking for, and therefore we didn't get the expected list. Unfortunately I won't have time to look into this in more details before Sunday/Monday. Just wanted to ask you to double check we don't revert that fix

Yes. Thoroughly tested. Now we will add $album->{artist} first, and then iterate through the $album->{artists} array looking for additional "main-artists", filtering out the one we got from $album->{artist}. Very straightforward, and tested with AVB. The problem was that we were missing the primary artist in some cases because it was not also contained in the main-artist array.

michaelherger (approved):
Well, then I'll go ahead and approve this. Thanks!

PR #98, opened 2025-05-08
Add Qobuz album description to the "On Qobuz" menu

SamInPgh:
This can provide useful background information and editorial opinion regarding an album, especially for those of note.


michaelherger (approved):
LGTM

PR #103, opened 2025-06-21
Slim browse metadata

darrell-k:
From darrellk/SqueezeboxQobuz (branch SlimBrowse-Metadata)

Work in progress - changes to add item type/metadata to SlimBrowse responses.

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1249,11 +1249,11 @@ sub QobuzGetTracks {
 
 			$totalDuration += $track->{duration};
 			my $formattedTrack = _trackItem($client, $track);
-			my $work = delete $formattedTrack->{work};
+			my $work = $formattedTrack->{work};
 
 			# create a playlist for each "disc" in a multi-disc set except if we've got works (mixing disc & work playlists would go horribly wrong or at least be confusing!)
 			if ( $prefs->get('showDiscs') && $formattedTrack->{media_count} > 1 && !$work ) {
-				my $discId = delete $formattedTrack->{media_number};
+				my $discId = $formattedTrack->{media_number};
 				$discs->{$discId} = {
 					index => $trackNumber,
 					title => string('DISC') . " " . $discId,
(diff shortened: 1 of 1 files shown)


michaelherger:
Wouldn't this change remove some menu items from non-Material UIs?

darrell-k:
> Wouldn't this change remove some menu items from non-Material UIs?

No, the existing menu items would only be suppressed if the `tags` parameter is defined.

Having said that, I've found another problem: the changes are somehow messing up the track title text returned to Default Skin. I'm looking at this.

Plugin.pm
```diff
@@ -1746,10 +1753,20 @@ sub _trackItem {
 	my $genre = $track->{album}->{genre};
 
 	my $item = {
-		name  => sprintf('%s %s %s %s %s', $title, cstring($client, 'BY'), $artist, cstring($client, 'FROM'), $album),
+		name  => $tags ? $title : sprintf('%s %s %s %s %s', $title, cstring($client, 'BY'), $artist, cstring($client, 'FROM'), $album),
```
michaelherger, line 1756: I think you should leave the `name` alone and add a `title` element instead. `XMLBrowser` would then return that raw title to the calling client. At least that's what I've done in my Bandcamp modification. Please give it a try and let me know about the outcome.

michaelherger: See https://github.com/LMS-Community/slimserver/blob/SlimBrowse-Metadata/SlimBrowse%20Metadata.md

darrell-k: The problem is here:

https://github.com/LMS-Community/slimserver/blob/65ced03a17bfd6376f84a71c4f689892c062e60c/Slim/Web/XMLBrowser.pm#L830-L837

when `$_->{'hasMetadata'}` is passed, `$_->{'name'}`  is set via TitleFormatter, but the existing plugin code doesn't send `$_->{'title'}` so TitleFormatter doesn't return anything useful.

The assumption up to now is that the plugin sets `$_->{'name'}` directly:
`sprintf('%s %s %s %s %s', $title, cstring($client, 'BY'), $artist, cstring($client, 'FROM'), $album)`
which is not to be manipulated by `Slim::Web::XMLBrowser`. But passing `$_->{'hasMetadata'}` breaks this.

So the problem isn't with `$_->{'name'}` but the absence of `$_->{'title'}`.

I think the best solution is to always pass `$_->{'title'}` back from the plugin so that the titleFormatWeb preference is followed and  `$_->{'name'}` as set in the plugin is ignored.

I'll also have to check what happens with `Slim::Buttons::XMLBrowser`.

What do you think?

michaelherger: Exactly: you need to provide the "raw" title.

PR #104, opened 2025-06-22
Always set 'title' element

darrell-k:
Something's happening in XMLBrowser which messes up the title format when the extra metadata is passed back, so only return it when we've got the `$tags` parameter.

Maybe we should look at XMLBrowser, in case we ever wanted to use `$tags` in Default Skin requests?

### DIFF
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -1703,7 +1703,6 @@ sub _albumItem {
 		album_id => $album->{id},
 		album_title => $album->{title},
 	}];
-$item->{version}= 'darrell album';
 
 	return $item;
 }
@@ -1841,7 +1840,6 @@ sub _trackItem {
 	$item->{tracknum} = $track->{track_number};
 	$item->{media_number} = $track->{media_number};
 	$item->{media_count} = $track->{album}->{media_count};
-$item->{titleFlags}= 'darrell track';
 	return $item;
 }
 


darrell-k:
Changes as discussed in #103

Plugin.pm
```diff
@@ -1791,12 +1793,14 @@ sub _trackItem {
 			my @titleSplit = split /:\s*/, $track->{title};
 			my $tempTitle = @titleSplit[-1];
 			$item->{work} =~ s/:\s*\Q$tempTitle\E//;
+			$item->{title} =~ s/\Q$item->{work}\E://;
```
michaelherger, line 1796: As the `title` is supposed to be the raw title, shouldn't we leave it alone? How did the work get in there?

Plugin.pm
```diff
@@ -1807,6 +1811,7 @@ sub _trackItem {
 	}
 
 	if ( $prefs->get('parentalWarning') && $track->{parental_warning} ) {
+		$item->{title} .= ' [E]';
```
michaelherger, line 1814: Let's really try to leave the title alone. This should be the title and nothing else, as it might be used to look up additional metadata etc.

darrell-k:
Just added a commit to keep album name and track title clean by adding elements: `version` for albums and `titleFlags` for tracks.

Plugin.pm
michaelherger: Classical is too complicated for me. The music always was 😁.

Plugin.pm
michaelherger: I'd be in support of an extra element for such additions. As you might know I'm often struggling trying to match names and titles to whatever a web site might be using. And all these additions complicate things considerably.

michaelherger (approved):
LGTM

darrell-k:
> > Just added a commit to keep album name and track title clean by adding elements: `version` for albums and `titleFlags` for tracks.
> 
> I'm just curious as to why the "[E]" and "*" modifiers are added to the `version` element with a trailing space, e.g. "[E] ", rather than a leading space, e.g. " [E]", as was previously done with the `title` string. Typo, or am I missing something?

Because unlike name/title, the new element starts off empty.

PR #107, opened 2025-07-05
Further metadata enhancements, display and scan

darrell-k:
Further metadata enhancements:

1. Retired `API::Common::getArtistName` in favour of `API::Common::getMainArtists`
2. Use arrays for artist, artist_extid.
3. Do not use Qobuz "Various Artists" for ARTIST. If this would leave ARTIST empty, use track=>performer->name. This allows the LMS scanner to work its own VA magic and prevents "Various Artists" from appearing as a track-level artist.
4. Try to extract Orchestra and Conductor from Qobuz metadata for the scanner.
5. Plugin track listing items and library track listings for imported albums *should* now be consistent. 

This is WIP, I'm still working on improvements for classical music scanning, but pushing now to give people a chance to comment on the changes and to check that I've not broken anything.

Ignore the commented out code at the bottom of Importer.pm, that's for my continuing analysis.


SamInPgh:
Looks good, Darrell. I'm looking forward to testing. Will the changes to import ARTIST and ARTIST_EXTID as arrays require a full library rescan or will an online library rescan suffice?

darrell-k:
I think a full rescan is always best when fundamental scanner changes are implemented. And specifically relevant to this change is that I know that there are inconsistencies in a new & changed scan with regard to changes in VA/Compilation status. (There's an LMS PR on this, incomplete, I must get back to it!)

SamInPgh:
Please merge when ready and I will begin testing. I'm unclear on the status of the SlimBrowser changes though.

SamInPgh:
I was able to install the changes on my system and did some testing. Generally, things look very good. A couple of observations:

1. For tracks where a track artist exists, the track listing from the plugin now shows both the album artist(s) and the track artist whereas the library view shows only the track artist for those tracks. I personally prefer the latter approach as it seems redundant to show the album artist(s) there. It could be that there is a server setting somewhere that is responsible for this behavior on my system though. 
2. When showing a list of releases in the plugin, e.g. favorite releases, new releases, etc., only the name of the primary artist (album->artist->name) is displayed, whereas the library release view shows all album artist names.
3. When multiple album artists are displayed in Material's "Playing" and "Queue" views, clicking on any of the artist names seems to select the first artist only, whereas each of the artist names can be separately selected from the library release and track list views. (I suspect that this is a Material issue but wanted to point it out).

SamInPgh:
Here's an example of item 1 in my previous comment.

Plugin tracklist:

Screenshot_20250707-000857

Library tracklist:

Screenshot_20250707-000953

API/Common.pm
```diff
@@ -248,13 +239,22 @@ sub precacheTrack {
...
+	}
 
 	my $meta = {
 		title    => $track->{title} || $track->{id},
 		album    => $album->{title} || '',
 		albumId  => $album->{id},
-		artist   => $class->getArtistName($track, $album),
-		artistId => $album->{artist}->{id} || '',
+		artist   => join(', ', Slim::Utils::Misc::uniq(@artistNames)),
+		artistId => join(', ', Slim::Utils::Misc::uniq(@artistIds)),
```
michaelherger, line 257: I'm not sure how `List::Util::uniq` is working: does it always preserve the order in the list? Does it always remove from the end? Just a thought. Our own (fallback) implementation certainly does.

It might be safer to not add dupes in the first place - at the cost of a slight performance overhead.

Plugin.pm
```diff
@@ -1741,7 +1741,12 @@ sub _trackItem {
 	my ($client, $track, $isWeb) = @_;
 
 	my $title = Plugins::Qobuz::API::Common->addVersionToTitle($track);
-	my $artist = Plugins::Qobuz::API::Common->getArtistName($track, $track->{album});
+	my @artistNames = map { $_->{name} } Plugins::Qobuz::API::Common->getMainArtists($track->{album});
+	@artistNames = grep { !/^\s*VARIOUS\s*ARTISTS\s*$/i } @artistNames;
```
michaelherger, line 1745: Could this be subject to localisation? Is this a "tag style" value, or a literal? Qobuz is French. You might get surprises....

michaelherger:
> Please merge when ready and I will begin testing. I'm unclear on the status of the SlimBrowser changes though.

Are you referring to the "enhanced metadata" changes? I haven't seen much feedback on the work so far. Did I miss some of Craig's feedback?

API/Common.pm
darrell-k: `Slim::Utils::Misc::uniq` works well in Queries.pm (albumQuery), where it de-duplicates the array of roles which drives the artist display sequence. So I think this is fine as it is.

michaelherger: `S::U::Miscellaneous::uniq` would use a binary module - if installed in the OS, or our own fallback code. I don't know which one is used on your system. But it's probably reasonable to assume that deduplication would prefer those items early in a list.

darrell-k:
> 1. For tracks where a track artist exists, the track listing from the plugin now shows both the album artist(s) and the track artist whereas the library view shows only the track artist for those tracks. I personally prefer the latter approach as it seems redundant to show the album artist(s) there. It could be that there is a server setting somewhere that is responsible for this behavior on my system though.

There is code in LMS that always pushes the album artist (ie the artist in the `albums.contributor` database column) to the top of the list of contributors under each track when browsing the library. It's not a Qobuz thing, though maybe this is wrong, because you can now select/deselect any role in My Music (or Behaviour in Material) server settings.

> 2. When showing a list of releases in the plugin, e.g. favorite releases, new releases, etc., only the name of the primary artist (album->artist->name) is displayed, whereas the library release view shows all album artist names.

I'll look at this, although I'm grappling with a related question, which is: what do `album->artist` and `album->artists` in the Qobuz metadata actually signify? And is it consistent? (I think we already know the answer to that second question!)

My current experiment in the scanner is to
1. use `album->artist` as ALBUMARTIST rather than ARTIST (except for "Various Artists").
2. ignore `album->artists` as this can be amalgamation of all the "Main Artists" on the album, so is neither ALBUMARTIST, nor ARTIST nor TRACKARTIST in LMS terms. This is causing problems particularly with some box sets.
3. use `track->performer` as ARTIST rather than TRACKARTIST. There is some code in the LMS scanner which converts ARTIST to TRACKARTIST where necessary.

I'll let you know how it goes. I think it might be better, I'll test with your example albums as well as my own.
 
> 3. When multiple album artists are displayed in Material's "Playing" and "Queue" views, clicking on any of the artist names seems to select the first artist only, whereas each of the artist names can be separately selected from the library release and track list views. (I suspect that this is a Material issue but wanted to point it out).

This should be fixed in my https://github.com/darrell-k/slimserver/tree/music-service-metadata LMS branch. I'm going to create a PR for this today.

Plugin.pm
darrell-k: I changed my Qobuz interface language to French, and we're still getting "Various Artists" in English (in the Qobuz App, also), if that's any indication.

SamInPgh:
I'm not sure exactly what happened but, since updating to the latest nightly build of LMS 9.1.0 (Version: 9.1.0 - 1752002406 @ Wed Jul 9 03:13:46 UTC 2025), LMS goes into a tight error loop when trying to play any Qobuz content, with the following error message logged every .07 seconds:

```
[25-07-09 15:29:15.1379] Slim::Networking::IO::Select::__ANON__ (131) Error: Select task failed calling Slim::Networking::Async::HTTP::_http_read_body: Can't locate object method "processors" via package "Slim::Schema::Track" at /usr/share/perl5/Slim/Utils/Scanner/Remote.pm line 790.
; fh=Slim::Networking::Async::Socket::HTTPS=GLOB(0xa5689e8)

```
Am I missing some updates?

michaelherger:
> ```
> [25-07-09 15:29:15.1379] Slim::Networking::IO::Select::__ANON__ (131) Error: Select task failed calling Slim::Networking::Async::HTTP::_http_read_body: Can't locate object method "processors" via package "Slim::Schema::Track" at /usr/share/perl5/Slim/Utils/Scanner/Remote.pm line 790.
> ; fh=Slim::Networking::Async::Socket::HTTPS=GLOB(0xa5689e8)
> ```


This might be related to the latest changes in LMS which try to treat in-library remote tracks like local tracks. There must be some mix-up. Can you confirm whether this is a track you have imported in to your library or not?

SamInPgh:
> > ```
> > [25-07-09 15:29:15.1379] Slim::Networking::IO::Select::__ANON__ (131) Error: Select task failed calling Slim::Networking::Async::HTTP::_http_read_body: Can't locate object method "processors" via package "Slim::Schema::Track" at /usr/share/perl5/Slim/Utils/Scanner/Remote.pm line 790.
> > ; fh=Slim::Networking::Async::Socket::HTTPS=GLOB(0xa5689e8)
> > ```
> 
> This might be related to the latest changes in LMS which try to treat in-library remote tracks like local tracks. There must be some mix-up. Can you confirm whether this is a track you have imported in to your library or not?

Ahh! As you suspected, it only occurs with Qobuz tracks that have been imported into my library --- even when playing them via the plugin view rather than directly from the library.  Unimported Qobuz tracks play without a problem.

Importer.pm
```diff
@@ -348,28 +334,73 @@ sub _prepareTrack {
...
+	@artists = @$artists_ref;
+	@artistIds = @$artistIds_ref;
+
+	if ($track->{performer} && Plugins::Qobuz::API::Common->trackPerformerIsMainArtist($track) && !grep $_ eq $track->{performer}->{name}, @artists) {
+		push @artists, $track->{performer}->{name};
+		push @artistIds, 'qobuz:artist:' . $track->{performer}->{id};
+	}
+
+	$attributes->{ARTIST} = \@artists;
+	$attributes->{ARTIST_EXTID} = \@artistIds;
```
michaelherger, line 378: Oh, so this is what `Use arrays for artist, artist_extid.` in the description referred to? That's only happening in the browser?

I don't remember: do we support multiple `extid` values in the scanner?

API/Common.pm
darrell-k: Interesting. I've experimented and we don't pass through `artists`, only `artist`, in the status query. And (probably because there are no artist_ids passed) Neither Material nor Default Skin creates a link for the name in the playlist. However, passing multiples (with or without a space after the comma) does mess up the "More Info" options (again both skins) because the whole string is used for the lookups so MAI etc results.

So I think I'll revert here to a single artist only (which would be the main album level artist, unless "various artists", in which case we'd pass the primary track level artist).

For the plugin track listing (see change to Plugin.pm below) the contructed `$artist` string is only used as `line2` text, which is not used in any hyperlinks, so I propose leaving the new multiple artist string in place.

I think this is good enough for now, we can further enhance with the Slimbrowse enhancements. My current thinking there is to effectively retire the plugin track listing (for plugins that can handle it) in favour of a library browse listing, even for unimported albums.

Importer.pm
darrell-k: Yes, I tested `extid ` arrays and it works.

SamInPgh:
As I feared from reading the most recent review comments, my previous change to show all album artists in '$album->{artists}' equally in the library album listing and in the 'album artist' views has been lost. I'm not happy at all with what I'm seeing. My test case is "As Falls Wichita, So Falls Wichita Falls" by Pat Metheney and Lyle Mays. It is a truly equal collaboration, and both artists are in '$album->{artists}'. Not only is Lyle Mays no longer shown as an album artist, i.e. "Pat Matheney, Lyle Mays", in the album listing, he is also not to be found under the "Album Artists" view. This behavior is completely unacceptable to me. These last two commits need to be rethought, in my opinion.

darrell-k:
Let me have a look at that, Sam. It wasn't intentional.

SamInPgh:
Okay. I was hoping that was the case. I just noticed that both artists still appear in "Now Playing" and that MAI Bios are displayed for both, so that is a good sign...

Plugin.pm
```diff
@@ -1741,7 +1741,13 @@ sub _trackItem {
 	my ($client, $track, $isWeb) = @_;
 
 	my $title = Plugins::Qobuz::API::Common->addVersionToTitle($track);
-	my $artist = Plugins::Qobuz::API::Common->getArtistName($track, $track->{album});
+	my @artistNames = map { $_->{name} } Plugins::Qobuz::API::Common->getMainArtists($track->{album});
+	my ($artists_ref) = Plugins::Qobuz::API::Common->removeArtistsIfNotOnTrack($track, \@artistNames);
+	@artistNames = @$artists_ref;
+	if ($track->{performer} && Plugins::Qobuz::API::Common->trackPerformerIsMainArtist($track) ) {
+		push @artistNames, $track->{performer}->{name};
+	}
```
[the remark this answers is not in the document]
darrell-k: This started as inline code. Then I struggled with passing/returning multiple arrays after moving the code into a function. What you're suggesting makes much more sense.

SamInPgh:
> @SamInPgh For some reason I though that setting an ALBUMARTIST tag was a good idea. Before I push the change I'd like to try to remember why, or at least run my full tests without doing so.
> 
> In the meantime you could delete lines 380-383 in Importer.pm so you can carry on testing.

Thanks, @darrell-k . That took care of the problem. Here are some anomalies I have noticed so far, all having to do with reclassifying certain albums as "Various Artists" that were previously attributed to individual artists:

1. The Neil Young "Decade" album. AFAIK, there is only one "$album->{artist}" --- Neil Young. Although a few of the tracks were originally recorded by and credited to his previous bands, such as "Buffalo Springfield" and "Crosby, Stills, Nash and Young", this doesn't seem sufficient to justify reclassifying it.
2. "Blues Breakers" by John Mayall & The Bluesbreakers and Eric Clapton. Both are listed in "$album->{artist/s}" and should, in my opinion, still be co-credited as the album artist. Definitely not "Various Artists".
3. "The Four Quarters" by Solem Quartet. Again, there are other artists on some of the tracks but I don't see that as justification for reclassifying the whole album based on that.
4. "Super Session" has also been reclassified as "Various Artists". In this case, I have mixed feelings due to the inconsistency of the track artist tags.

Generally speaking, I don't believe that an album with specific "$album->{artist/s}" defined should be reclassified to "Various Artitsts" except under certain very specific conditions. One that I could agree with would be "The Metallica Blacklist", which has Metallica as the sole "$album->{artist}", but where the songs are all covers performed by other artists.

darrell-k:
The various artists problem is exactly what I've been working on: that was the reason why I was adding an ALBUMARTIST tag with the value of the primary artist (ie album->artist). But as you reported, that screwed up the album with two album artists.

The scanner notices that not all the artists are the same, and therefore makes it a compilation if there is no ALBUMARTIST tag. So what I'm doing now is keeping track of the artist tags created for each track on the album and if they vary track to track, I work out the common artists across the tracks and add those as ALBUMARTIST. That should mean that only albums without any common artist are set as compilations by the scanner.

Full testing will not be tonight, my brain is fried working out a way to do this semi-efficiently, and I still have make the technical changes requested by Michael.

darrell-k:
PS: On this basis, the Metallica album will not be classed as Various Artists, because I'm deliberately keeping the main `album->artist` unconditionally too avoid removing all artists in cases the track->performers string is really ****ed up. I'll have a think about whether there is another solution to that.

SamInPgh:
I don't have a problem with keeping the Metallica album the way that Qobuz tagged it. My feeling is that an album with one or more main artists tagged should only be reclassified as a compilation under very strict circumstances, it at all.

darrell-k:
I've a feeling that the way I've got things now, only albums that Qobuz tags with album->artist="Various Artists" and with differing artists across the tracks will be set as compilations.

API/Common.pm
```diff
@@ -248,13 +239,25 @@ sub precacheTrack {
...
+	foreach ( $class->getMainArtists($album) ) {
+		push @artistNames, $_->{name};
+		push @artistIds, $_->{id};
+	}
+	my ($artists_ref, $artistIds_ref) = Plugins::Qobuz::API::Common->removeArtistsIfNotOnTrack($track, \@artistNames, \@artistIds);
+	@artistNames = @$artists_ref;
+	@artistIds = @$artistIds_ref;
+	if ($track->{performer} && $class->trackPerformerIsMainArtist($track) ) {
+		push @artistNames, $track->{performer}->{name};
+		push @artistIds, $track->{performer}->{id};
```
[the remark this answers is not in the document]
darrell-k: @michaelherger , a question: does it make any difference, efficiency-wise to nevertheless return the reference? I must admit I prefer to see explicitly in code what is being asked for/returned. Maybe I don't have a full Perly brain yet.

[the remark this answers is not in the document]
michaelherger: I must admit that I don't know. But sometimes the efficiency loss is so little, it's worth the sacrifice in favour of readability. I would probably do the same. With a little luck Perl would recognise the situation and make the assignment of a reference to itself a no-op anyway.

But don't go the full reference/de-reference/reference cycle. That certainly has an impact as you'd be creating copies of the lists.

darrell-k:
Changes pushed. Apart from the reference/dereference changes as requested, I've added a new procedure to add ALBUMARTIST tags if required. This prevents the scanner setting the compilation flag whenever the artists across the album differ between tracks. The previous version simply set ALBUMARTIST=<main artist> but this failed @SamInPgh 's test above, for an album with two equal album artists. So:

1. I populate a hash `%allArtists` with the unique combinations of artist/artist-ids found while looping through the tracks of the album.
2. `_addAlbumArtists` is then called once all `$attributes` has been created for the album, but before `storeTracks`, to add ALBUMARTIST and ALBUMARTIST_EXTID tags if necessary. The logic uses `%allArtists` to find the artists which are used on every track and adds the tags for each one. If no common artists are found, no ALBUMARTIST tags are added so the scanner will set the album as a compilation in the normal way if ARTIST differs track to track.

Please give this a good test.

SamInPgh:
Okay. That's more than I was expecting but you can be sure that I will test it fully. 😉

Importer.pm
```diff
@@ -348,29 +339,92 @@ sub _prepareTrack {
...
+			my $name = shift @roles;
+			foreach (@roles) {
+				$_ =~ s/[\r$|\s*]//g;
+				$_ = uc($_);
+				push @{$rolePerformer->{$_}}, $name if !$roleSeen{$_};
+				$roleSeen{$_} = 1;
+			}
+		}
+		$attributes->{BAND} = $rolePerformer->{ORCHESTRA} if $rolePerformer->{ORCHESTRA};
+		$attributes->{CONDUCTOR} = $rolePerformer->{CONDUCTOR} if $rolePerformer->{CONDUCTOR};
```
michaelherger, line 400: Use named variables inside nested loops, rather than the anonymous `$_`. This is prone to errors, as it changes, depending on the context. Do `foreach my $performarAndRoles (@performersAndRoles)` etc.

Importer.pm
```diff
@@ -348,29 +339,92 @@ sub _prepareTrack {
...
+	my $rolePerformer;
+	if ($track->{performers}) {
+		my %seen;
+		my @performersAndRoles = split(' - ', $track->{performers});
+		foreach (@performersAndRoles) {
+			my %roleSeen = undef;
+			my @roles = split(/\s*,\s*/, $_);
+			my $name = shift @roles;
+			foreach (@roles) {
+				$_ =~ s/[\r$|\s*]//g;
```
michaelherger, line 393: I think `\r` (as well as `\n`) would be covered by `\s` if if you used the `s` modifier in the regex, wouldn't it?

Importer.pm
```diff
@@ -348,29 +339,92 @@ sub _prepareTrack {
...
+			}
+		}
+		$attributes->{BAND} = $rolePerformer->{ORCHESTRA} if $rolePerformer->{ORCHESTRA};
+		$attributes->{CONDUCTOR} = $rolePerformer->{CONDUCTOR} if $rolePerformer->{CONDUCTOR};
+	}
+
 	return $attributes;
 }
 
+sub _addAlbumArtists {
```
michaelherger, line 406: I have no clue what this method is supposed to do. Can you elaborate?

Plugin.pm
```diff
@@ -1741,7 +1741,13 @@ sub _trackItem {
 	my ($client, $track, $isWeb) = @_;
 
 	my $title = Plugins::Qobuz::API::Common->addVersionToTitle($track);
-	my $artist = Plugins::Qobuz::API::Common->getArtistName($track, $track->{album});
+	my $artistNames;
+	@{$artistNames} = map { $_->{name} } Plugins::Qobuz::API::Common->getMainArtists($track->{album});
```
michaelherger, line 1745:
```perl
my $artistNames = [ map { $_->{name} } Plugins::Qobuz::API::Common->getMainArtists($track->{album}) ];
```

Importer.pm
michaelherger: Ok, thanks. I think i got it.

Overall I believe this could be simplified a lot. In particular this should work without a global variable. I'll comment on the calling code.

SamInPgh:
I'm planning on testing with the latest commits today. Are there any more changes imminent?

darrell-k:
> I'm planning on testing with the latest commits today. Are there any more changes imminent?

Only technical ones, so you can still validate the behaviour.

Importer.pm
```diff
@@ -104,9 +106,11 @@ sub scanAlbums {
...
 			if ($albumDetails && ref $albumDetails && $albumDetails->{tracks} && ref $albumDetails->{tracks} && $albumDetails->{tracks}->{items}) {
 				$progress->update($album->{title});
-				$class->storeTracks([
-					map { _prepareTrack($albumDetails, $_) } @{ $albumDetails->{tracks}->{items} }
-				], undef, $accountName);
+
+				%allArtists = ();
+				my $attributes = [map { _prepareTrack($albumDetails, $_) } @{ $albumDetails->{tracks}->{items} }];
+				_addAlbumArtists($attributes);
+				$class->storeTracks($attributes, undef, $accountName);
```
michaelherger, line 113: It's my understanding that if no artist was represented on every single track we would not define an album artist, right?

As mentioned in `_addAlbumArtists()` I believe this could be simplified. This is not working code, but I hope you get the idea:

```perl
# these will be list refs once initialised, but must not be initialised here!
my $albumArtists;
my $albumArtistsIds;

my $attributes = [ map {
   my $attrs = _prepareTrack($albumDetails, $_);

   # create a map of artist id -> artist name tuples for the current item
   my %trackArtists;
   for (my $i = 0; $i < scalar @{$attrs->{ARTIST_EXTID}}; $i++) {
      $trackArtists{$attrs->{ARTIST_EXTID}->[$i]} = $attrs->{ARTIST}->[$i];
   }

   # if this is the first track, all track artists are potential album artists
   if (!$albumArtists) {
      $albumArtists = [ values %trackArtists ];
      $albumArtistsIds = [ keys %trackArtists ];
   }
   # not the first track
   else {
      # we are only interested in the artists we have seen before - anything else is not considered an album artist
      for (my $i = 0; $i < scalar @$albumArtists; $i++) {
         if (!$trackArtists{$albumArtistsIds->[$i]) {
            $albumArtists->[$i] = undef;
            $albumArtistIds->[$i] = undef;
         }
         # instead of fiddling with the index above I decided to wipe the value - now we have to remove empty values
         $albumArtists = grep { $_ } @$albumArtists;
         $albumArtistIds = grep { $_ } @$albumArtistIds;
      }
   }

   # this part needs good verification: I believe that as we're dealing with list refs, the referencing 
   # item should always be referencing the latest and greatest, even if it's changed after the 
   # assignment here:
   $attrs->{ALBUMARTIST} = $albumArtists;
   $attrs->{ALBUMARTIST} = $albumArtists;

   # now let's return this track's attributes to be collected outside the map
   $attrs;
} @{ $albumDetails->{tracks}->{items} }];
```

None of this has been tested! But it hopefully gives you an idea. It should replace the global variable, the `my $attributes = map....` line, plus hopefully `_addAlbumArtists`. The main advantage I'm seeing is that we're not building a large data structure (`%allArtists`) to then post-process a lot more data than we need to. Because we already know that only artists found on the first track are potential album artists. We only have to deal with them and validate whether they're on each track. And with each artist that is not found on any one track the list of potential candidate gets shorter. In your example below we'd only ever have to do one check, as the first track only had one artist.

And no global variable is required. Global variables should be a very last resort.

Importer.pm
darrell-k: Yes, I was thinking as I documented it above, that we could probably just build a list of artists found on every track in `_prepareTrack`.

And pass a ref for the list into that, and then into `_addAlbumArtists` when we've finished processing the album, instead of using the global variable.

Still, I had fun learning how to flatten arrays and count occurrences!

Importer.pm
michaelherger: Just in case you were looking at the suggested change in email only, I'd suggest you go online, as I've found a few issues and fixed the comment. But I don't think you'd get email updates for those changes. Sorry about that.

darrell-k: See my reply just now. The only thing in the example is that it's not the first track (random return of the hash!)

But yes, just add the first track's artists to a list, then remove if they're not present on subsequent tracks. The only thing we need to be careful of is to have some flag that tells us if the artists are the same on all tracks - in that case we don't want to set album artist at all.

SamInPgh:
> > I'm planning on testing with the latest commits today. Are there any more changes imminent?
> 
> Only technical ones, so you can still validate the behaviour.

Okay. I'm still not clear on everything that's being done but things look better for multi-artist albums now. Keeping in mind my preference to stick with Qobuz's tags for album artist(s) whenever possible in order to match what is shown in their apps, I have found that one of my test cases no longer does so. It should look familiar. Here it is in the Qobuz app:

Screenshot_20250719-174104

Prior to the latest changes, it was displayed in the plugin and library views with all five album artists shown. Now it is not, even though all are listed in the plugin's tracklist view:

Screenshot_20250719-180134

I'm guessing that maybe it failed some sort of test based on the track artist contents but, as I've said before, I don't feel that we should be trying to correct or modify Qobuz's tagging choices unless they are typos that we can somehow detect. Anyway, I will continue testing.

P.S. I would still like to see all the album artists listed on the plugin's Releases page, as they are in the library views, if possible.

darrell-k:
> I'm guessing that maybe it failed some sort of test based on the track artist contents but, as I've said before, I don't feel that we should be trying to correct or modify Qobuz's tagging choices unless they are typos that we can somehow detect. Anyway, I will continue testing.
> 
> P.S. I would still like to see all the album artists listed on the plugin's Releases page, as they are in the library views, if possible.

On Jazz At The Pawnshop, I just went to have a look in the Qobuz app and there are two versions, with different metadata. The one I've favorited only has the one artist, Arne Domnerus, I'll try importing the other one which seems to be the one you are showing. 

On the releases listing, we're only showing album-artist-name which I think is all we've ever done.

darrell-k:
> > On Jazz At The Pawnshop, I just went to have a look in the Qobuz app and there are two versions, with different metadata. The one I've favorited only has the one artist, Arne Domnerus, I'll try importing the other one which seems to be the one you are showing.
> 
> Yes. I have both of them in my favorites.
> 

> > On the releases listing, we're only showing album-artist-name which I think is all we've ever done.
> 
> Correct. But I'd like to change that to match the library view, which now shows all of them.

Easy enough, I'll make the change.

On the track listing (taking the 2014 remastered version as the example), looks to me that we are doing exactly what Qobuz does: it doesn't list the extra artists at track level because they're not in the track level metadata.

SamInPgh:
> 
> Easy enough, I'll make the change.

Great.

> 
> On the track listing (taking the 2014 remastered version as the example), looks to me that we are doing exactly what Qobuz does: it doesn't list the extra artists at track level because they're not in the track level metadata.

I sent you the wrong screenshot for that one. I wasn't so much concerned about the track listing as the Release view, where it shows only Arne Domnerus as the album artist. They are all listed as '$album->{artists}' and are shown as such in the Qobuz app. (See the first screenshot in my previous post.) Therefore they all should at least be shown as album artists in the plugin and library Release views. There is an argument to be made for also showing them in the track listings, since they are all equal members of the five-piece combo, but I suppose that would again be a case of trying to compensate for Qobuz's tagging vagaries.

darrell-k:
> Going back to the other "Jazz At The Pawnshop" version, the 30th Anniversary 3-CD set attributed on Qobuz to "Various Artists", it is now shown in the library view under artist Arne Domnerus only with no mention of the other four artists even at the track level, even though all five of the artists are listed as "MainArtist" on every track. My expectation was that all the tracks on this release would be tagged with all five artist names. Why only Domnerus?

The other artists are only in track->performers. For artists, s far we've only been using that to check that track->performer is a "MainArtist". I could pull out other MainArtists from track->performers aswell.

darrell-k:
@michaelherger I couldn't make this part of your code work:

```
# this part needs good verification: I believe that as we're dealing with list refs, the referencing 
# item should always be referencing the latest and greatest, even if it's changed after the
# assignment here:
$attrs->{ALBUMARTIST_EXTIDS} = $albumArtistIds;
$attrs->{ALBUMARTIST} = $albumArtists;
```
So I have implemented a version of your approach in `_prepareTrack` and have had to keep `_addAlbumArtists` to post-process `$attributes` in a much simplified form.

Commit incoming.

darrell-k:
@SamInPgh , the only (intended!) functional change with this commit is to include all album-level artists in the plugin Releases listing, as you requested. I think I've tested well for regressions, but if you have a little time...

SamInPgh:
> @SamInPgh , the only (intended!) functional change with this commit is to include all album-level artists in the plugin Releases listing, as you requested. I think I've tested well for regressions, but if you have a little time...

Sorry for the delay. My brother was visiting from out of town today.

Good news:  All album-level artists are now shown in the plugin Releases listing, which is great.

Bad news:  Albums that have an album artist  of "Various Artists", such as one of the "Jazz At The Pawnshop" titles, show nothing at all in the artist field of the plugin Releases list, although they appear in the sorted list where "Various Artists" would be.

Screenshot_20250720-223804

michaelherger:
> @michaelherger I couldn't make this part of your code work:
> 
> 
> 
> ```
> 
> # this part needs good verification: I believe that as we're dealing with list refs, the referencing 
> 
> # item should always be referencing the latest and greatest, even if it's changed after the
> 
> # assignment here:
> 
> $attrs->{ALBUMARTIST_EXTIDS} = $albumArtistIds;
> 
> $attrs->{ALBUMARTIST} = $albumArtists;
> 
> ```

What I tried to say there was that as you are not storing the values for the album artist, but a reference to the list of them, you wouldn't have to post-process them. The referenced values will be updated as you process the list.

darrell-k:
Yes, I get the theory, I'll try again.

darrell-k:
> Bad news: Albums that have an album artist of "Various Artists", such as one of the "Jazz At The Pawnshop" titles, show nothing at all in the artist field of the plugin Releases list, although they appear in the sorted list where "Various Artists" would be.

Should be fixed in the latest commit.

SamInPgh:
> > Bad news: Albums that have an album artist of "Various Artists", such as one of the "Jazz At The Pawnshop" titles, show nothing at all in the artist field of the plugin Releases list, although they appear in the sorted list where "Various Artists" would be.
> 
> Should be fixed in the latest commit.

Great! Glad it was a simple fix. I have to say that I really like this change, not only for browsing my favorite releases but more so when browsing Qobuz lists such as New Releases or Bestsellers, where I can now easily spot interesting collaborations that I would not have otherwise been aware of without drilling down to the track list.

darrell-k:
> > @michaelherger I couldn't make this part of your code work:
> > ```
> > 
> > # this part needs good verification: I believe that as we're dealing with list refs, the referencing 
> > 
> > # item should always be referencing the latest and greatest, even if it's changed after the
> > 
> > # assignment here:
> > 
> > $attrs->{ALBUMARTIST_EXTIDS} = $albumArtistIds;
> > 
> > $attrs->{ALBUMARTIST} = $albumArtists;
> > ```
> 
> What I tried to say there was that as you are not storing the values for the album artist, but a reference to the list of them, you wouldn't have to post-process them. The referenced values will be updated as you process the list.

OK, the problem was your
```
$albumArtists = grep { $_ } @{$albumArtists};
```
was being evaluated in the scalar context, so `$albumArtist` ended up as 1, 2 or whatever and the code failed to run. Yesterday I changed that to:
```
$albumArtists = [ grep { $_ } @{$albumArtists} ];
```
but that resulted in a new reference in `$albumArtists` for each track. I just resolved this with:
```
@$albumArtists = grep { $_ } @{$albumArtists};
```
and your code now works.

However, we do not want to send an ALBUMARTIST tag at all if we don't identify a common artist(s) because sending an empty ALBUMARTIST tag messes up the scanner: even setting to undef prevents the scanner from creating a Various Artists compilation as we need it to if there are no common artists. We just end up with an ARTIST of "No Artist" and COMPILATION=0.

In other words, we don't know for sure that we want ALBUMARTIST tags at all until we've processed the last track on the album and setting the reference to `undef` is not enough, we need to reprocess `$attributes` to remove `$attributes->{ALBUMARTIST}` from each track.

We _could_ change `Slim::Schema::_mergeAndCreateContributors` to check for an empty array, but that would introduce a dependency on LMS 9.1 which I would rather avoid.

michaelherger:
Ok, got it. That post processing should still be much simpler than before. 

And we could fix the LMS side anyway.

SamInPgh:
> 
> @SamInPgh I'd be very grateful if you could run your import tests one more time, just in case!

My testing shows no difference from the previous version. Hopefully that's what you were expecting.

I like where things are now from a user perspective for the most part. Any remaining issues I have are primarily due to Qobuz tagging anomalies and inconsistencies, which unfortunately are common. As demonstrated by the recent case of the Alison Krauss "Arcadia" release, however, Qobuz apparently is willing to take action to correct obvious metadata errors when brought to their attention.

Here's an example of one that I wish we could somehow compensate for:  
the iconic "John Mayall and the Bluesbreakers With Eric Clapton" album is tagged by Qobuz with both "JM & the BB" and "EC" tagged as Album Artists, with both also tagged as Track Performers on 22 of the 23 tracks. However, track 3 for unknown reasons is missing Eric in the Track Performers (he definitely plays on the track). For that reason, apparently, he is not imported into the library as an album artist and the album does not appear under his name in the Album Artists. Yes, it's a Qobuz tagging error but one that we could conceivably handle by giving weight to the fact that he's tagged as an Album Artist by Qobuz. It's a more extreme case of the "Jazz at the Pawnshop" release where all five members of the band are tagged as Album Artists but only one of them (the now infamous Arne Domnerus) is tagged as a Track Performer in each of the tracks and therefore becomes the only imported Album artist.

Having said all that, I'm very happy overall with these improvements, especially in the way that multiple album artists are now recognized and handled. Great job, @darrell-k !

API/Common.pm
```diff
@@ -343,6 +346,30 @@ sub trackPerformerIsMainArtist {
 	}
 }
 
+sub removeArtistsIfNotOnTrack {
+	my ($class, $track, $artists, $artistIds) = @_;
+
+	if ($artists && scalar @$artists) {
+		my ($mainArtist, $mainArtistId);
```
michaelherger, line 353: This line is useless, as you re-define the variables below.

Importer.pm
```diff
@@ -348,29 +345,111 @@ sub _prepareTrack {
...
+		}
+
+		# instead of fiddling with the index above I decided to wipe the value - now we have to remove empty values
+		@{$albumArtists->{names}} = grep { $_ } @{$albumArtists->{names}};
+		@{$albumArtists->{ids}} = grep { $_ } @{$albumArtists->{ids}};
+	}
+
+		$attributes->{ALBUMARTIST} = $albumArtists->{names};
+		$attributes->{ALBUMARTIST_EXTID} = $albumArtists->{ids};
+
```
michaelherger, line 439: Indentation

API/Common.pm
darrell-k: Ah yes. but presumably, if you declare a variable conditionally (eg `my $mainArtistId = $artistIds->[0] if $artistIds;`) you might run into problems later because it might not be declared?

I think I meant:
```
		my ($mainArtist, $mainArtistId);
		$mainArtist = $artists->[0];
		$mainArtistId = $artistIds->[0] if $artistIds;
```

darrell-k:
> Here's an example of one that I wish we could somehow compensate for: the iconic "John Mayall and the Bluesbreakers With Eric Clapton" album is tagged by Qobuz with both "JM & the BB" and "EC" tagged as Album Artists, with both also tagged as Track Performers on 22 of the 23 tracks. However, track 3 for unknown reasons is missing Eric in the Track Performers (he definitely plays on the track). For that reason, apparently, he is not imported into the library as an album artist and the album does not appear under his name in the Album Artists. Yes, it's a Qobuz tagging error but one that we could conceivably handle by giving weight to the fact that he's tagged as an Album Artist by Qobuz. It's a more extreme case of the "Jazz at the Pawnshop" release where all five members of the band are tagged as Album Artists but only one of them (the now infamous Arne Domnerus) is tagged as a Track Performer in each of the tracks and therefore becomes the only imported Album artist.

On the basis that so far we've only identified superfluous album artists tagged in classical releases, we could limit that check for classical albums only, for now. What do you think?

SamInPgh:
One remaining inconsistency with multiple album artists that I have noticed:
When playing a multi-artist release that is not in the library via the plugin, only the first artist is shown in Now Playing and the Queue, as well as in the Artists tab of MAI. It would be great if all the main artists could be shown, as is the case for imported releases.

darrell-k:
> One remaining inconsistency with multiple album artists that I have noticed: When playing a multi-artist release that is not in the library via the plugin, only the first artist is shown in Now Playing and the Queue, as well as in the outArtists tab of MAI. It would be great if all the main artists could be shown, as is the case for imported releases.

See https://github.com/LMS-Community/plugin-Qobuz/pull/107#discussion_r2210936621 above. (I know this is a very long discussion, I'd completely forgotten about this until looking at the status query response reminded me!)

The coming Slimbrowse changes will sort this out (somehow!) though there will need to be a further change to the plugin as well.

API/Common.pm
michaelherger: The conditional assignment would define the variable with an undefined value if the condition failed.

API/Common.pm
```diff
@@ -343,6 +346,30 @@ sub trackPerformerIsMainArtist {
 	}
 }
 
+sub removeArtistsIfNotOnTrack {
+	my ($class, $track, $artists, $artistIds) = @_;
+
+	if ( ( !$track->{album}->{genres_list} || grep(/Classique/,@{$track->{album}->{genres_list}}) ) && $artists && scalar @$artists ) {
```
michaelherger, line 352: Any word about the genre check? Why only do this if it was classical music or no genre?

API/Common.pm
darrell-k: So a bad assumption on my part.

API/Common.pm
darrell-k: A couple of replies up. We've only seen (so far) extraneous artists at the album level in classical releases.

On the other hand, we keep finding rock/jazz/blues collaborations where the artists are listed at the album level, but not on every/any track, even though we know they're playing on them.

SamInPgh: So is anyone else ready for a merge and release? I'm interested in getting some feedback from the user base.

PR #109, opened 2025-07-29
do not use Slim::Utils::Misc::uniq

darrell-k:
It's not compatible with LMS 8.5!


SamInPgh:
Looks good to me. Publishing.

darrell-k:
Thanks. 

Note to self: always test with LMS 8.5!

PR #112, opened 2025-08-07
Password encode UTF-8

darrell-k:
A user reported login failure: it turned out to be a "Wide Character" error because their password included the `€` symbol.

@michaelherger is this change fine with ancient Perl versions that might be in use?

### DIFF
--- a/Settings.pm
+++ b/Settings.pm
@@ -65,7 +65,7 @@ sub handler {
 
 		if ($params->{'username'} && $params->{'password'}) {
 			my $username = $params->{'username'};
-			my $password = md5_hex($params->{'password'});
+			my $password = md5_hex(Encode::encode("UTF-8", $params->{'password'}));
 
 			Plugins::Qobuz::API->login($username, $password, sub {
 				my $token = shift;


michaelherger:
> A user reported login failure: it turned out to be a "Wide Character" error because their password included the `€` symbol.
> 
> @michaelherger is this change fine with ancient Perl versions that might be in use?

I don't know, TBH. But I'd be positive. Except that I'm not sure whether we're good _always_ encoding. This stuff has always confused me, as we sometimes need to encode, sometimes we don't, depending on the platform LMS is running on. That said this might be limited to data read from the file system rather than user input like a password. 

I'd say go for it. It's simple enough to back out if needed. And I'm sure you tested with your own password?

darrell-k:
Tested with my own account (with a € symbol in the password) on:

Perl Version: 5.14.1 - MSWin32-x86-multi-thread
Perl Version: 5.32.1 - MSWin32-x64-multi-thread
Perl Version: 5.36.0 - aarch64-linux-gnu-thread-multi
Perl Version: 5.36.0 - aarch64-linux-gnu-thread-multi (Android/Termux)
Perl Version: 5.36.0 - x86_64-linux-gnu-thread-multi

darrell-k:
Thanks.

PR #113, opened 2025-10-13
Populate track composer tag from track performers

SamInPgh:
Use track->Performers to detect additional Composer credits when available.

### DIFF
--- a/Importer.pm
+++ b/Importer.pm
@@ -384,7 +384,7 @@ sub _prepareTrack {
 		my $name = $track->{composer}->{name};
 		$name =~ s/[\s\x{A0}]+/ /g;  # convert white space to a single space
 		push @$composers, $name;
-		$seen{$name} = 1;  # add it to the seen composers hash table
+		$seen{uc($name)} = 1;  # add it to the seen composers hash table
 		push @$composerIds, 'qobuz:artist:' . $track->{composer}->{id};
 		if ( $track->{work} && $prefs->get('importWorks') ) {
 			$attributes->{WORK} = $track->{work};
@@ -411,10 +411,10 @@ sub _prepareTrack {
 		# Populate secondary composers if any
 		if (!$track->{work}) { # works only allow one composer
 			if ($rolePerformer->{COMPOSER}) {
-				push @$composers, grep { !$seen{$_}++ } @{$rolePerformer->{COMPOSER}};
+				push @$composers, grep { !$seen{uc($_)}++ } @{$rolePerformer->{COMPOSER}};
 			}
 			elsif ($rolePerformer->{WRITER}) {  # if no track composers, use track writers
-				push @$composers, grep { !$seen{$_}++ } @{$rolePerformer->{WRITER}};
+				push @$composers, grep { !$seen{uc($_)}++ } @{$rolePerformer->{WRITER}};
 			}
 		}
 


Importer.pm
```diff
@@ -383,9 +375,22 @@ sub _prepareTrack {
 		push @$artistIds, 'qobuz:artist:' . $track->{performer}->{id};
 	}
 
-	my $rolePerformer;
+	$attributes->{ARTIST} = \@$artists;
+	$attributes->{ARTIST_EXTID} = \@$artistIds;
```
michaelherger: Assuming `$artists` was a list reference, then prefixing it with `@$artists` would de-reference it, and `\@$artists` would reference the de-referenced list - resulting in the same again? Or is this a trick to clone the list?

SamInPgh:
> Looks like the difference between a standard space and a non-breaking space:

I think the `grep` can be modified to handle that. Right?

darrell-k:
> So now we're back to how the arrays are referenced/dereferenced. Is that actually an issue? I mean, it seems to work... 😀

Should just be a case of removing `\@` from the following lines:
```
$attributes->{COMPOSER} = \@$composers;
$attributes->{COMPOSER_EXTID} = \@$composerIds;
$attributes->{ALBUMARTIST} = \@$albumArtists->{names};
$attributes->{ALBUMARTIST_EXTID} = \@$albumArtists->{ids};
```
On case differences, yes it's a Qobuz tagging issue but then isn't everything? I bet the non-breaking space came from someone copy/pasting from a web page!

SamInPgh:
> Should just be a case of removing `\@` from the following lines:
> 
> ```
> $attributes->{COMPOSER} = \@$composers;
> $attributes->{COMPOSER_EXTID} = \@$composerIds;
> $attributes->{ALBUMARTIST} = \@$albumArtists->{names};
> $attributes->{ALBUMARTIST_EXTID} = \@$albumArtists->{ids};
> ```

How about these lines?

	$attributes->{ARTIST} = \@$artists;
	$attributes->{ARTIST_EXTID} = \@$artistIds;

darrell-k:
> > Oh... I confused it with `0x0A`... I'd then recommend using a class `s/[\s\x{A0}]+/ /g`. Would that work?
> 
> It could. My worry is that going any further right now based on theoretically possible circumstances might backfire, resulting in unanticipated problems. I'd feel more comfortable proceeding with the fix for this specific known condition right now and taking care of any possible others as they present themselves.

That enhancement doesn't look like it could cause problems, unless there are circumstances where we wouldn't want to replace whitespace with a single space. I can't think of any.

SamInPgh:
@mherger  I  thought you might be interested in this error I got from the MAI plugin while doing an online only rescan of my library. It's happened twice in a row now.

`[25-10-15 17:31:09.5480] main::main (352) Error: Failed when running scan post-process: [Can't use string ("read timeout at /usr/share/squee"...) as a HASH ref while "strict refs" in use at /var/lib/squeezeboxserver/cache/InstalledPlugins/Plugins/MusicArtistInfo/Importer.pm line 310.
`

darrell-k:
Sorry, I forgot to test this earlier. Have done so now and it looks great. For your next task, write something that can recognise that Shostakovich and Chostakovich are the same composer...

SamInPgh:
> For your next task, write something that can recognise that Shostakovich and Chostakovich are the same composer...

Ha ha!! I think that code would go after the `else` portion of the `if (!$track->{work}) {` code block. Seriously though, there are a lot of variations on composer names in that track performers data --- taking Paul McCartney as an example, in my Composers table we have Paul James McCartney, Paul McCartney and Paul Mccartney. For Bob Dylan, there's also B. Dylan. Maybe AI can figure it out for us.

SamInPgh:
> Heh... while testing TIDAL yesterday I went to the credits page for the latest 3.0 album. He was named Gordon Sumner on one single track of that album.

The same on Qobuz. He is listed as the composer on the last track of the album only. That tells me it's not the fault of the streaming services. They're just reporting what they are receiving from whoever it is that supplies it.

PR #127, opened 2026-03-31
Favorites management API changes

darrell-k:
Replace favorite/getUserFavorites with  favorite/getUserFavoriteIds or favorite/status when retrieving the status of favorites. It's a lot less data.

In `QobuzGetTracks` and `QobuzArtist` use favorite/status to return a simple boolean.

In `QobuzManageFavorites` use favorite/getUserFavoriteIds because although it's more data to return than favorite/status, it avoids having to call the status API 3 times (for albums, tracks and artists).

The new API results are cached for 60 seconds, like favorite/getUserFavorites, so extend the cache management in `createFavorite` and `deleteFavorite`.

Please give this a good going over.


SamInPgh:
Running with the changes now. Do you anticipate this having any affect on the OMLI import process? I haven't looked at the code yet but could imagine a possible conflict.

darrell-k:
There shouldn't be any effect on OMLI, but worth checking.

API.pm
```diff
@@ -408,7 +438,11 @@ sub createFavorite {
 
 	$self->_get('favorite/create', sub {
 		$cb->(shift);
-		$self->getUserFavorites(sub{}, 'refresh')
+		$self->getUserFavorites(sub{}, 'refresh');
+		$self->getUserFavoriteIds(sub{}, 'refresh');
+		$self->getUserFavoriteStatus(sub{}, { type => 'album', item_id => $args->{album_ids} }, 'refresh') if $args->{album_ids};
+		$self->getUserFavoriteStatus(sub{}, { type => 'artist', item_id => $args->{artist_ids} }, 'refresh') if $args->{artist_ids};
+		$self->getUserFavoriteStatus(sub{}, { type => 'track', item_id => $args->{track_ids} }, 'refresh') if $args->{track_ids};
```
michaelherger: This is doing more work than before. Is it worth it? Wouldn't the first call have all the information the subsequent two would get us?

darrell-k: We're just clearing caches here, after an add or delete favourite. Maybe we should do it within the add/delete api routines. At least we would know which of the 3 types to clear in the case of the status cache.

michaelherger: This is not only wiping caches, you're doing calls to eg. `favorite/status`, too, right?

darrell-k: Yes, sorry I meant wipe/recreate cache. We could delay the recreation of the cache entry until it is requested again, so just do the clear at that stage.

darrell-k:
See new commit. I don't think it's worth refreshing the favorites caches when favourites change when they only last 60 seconds anyway, so we just delete them.

API.pm
```diff
@@ -756,8 +756,13 @@ sub _get {
 
 	if ($params->{_wipecache}) {
 		$cache->remove($cacheKey);
+		if ( $params->{_wipecache} eq 'delete_cache_no_refresh' ) {
```
michaelherger: As we are dealing with named parameters in `$params`, why not pass `$params->{delete_cache_no_refresh}` instead of doing a string comparison, rendering `_wipecache` ambiguous? At the very least I'd strongly recommend you replace the repeated use of the string literal with a constant. That way a typo would immediately cause the start to fail, rather than fail the comparison at run time, leading to unexpected behaviour.

darrell-k: See latest commit. I've changed the favorites APIs to use only named parameters. I've also allowed for callers to override any parameters defined in the individual APIs if they want to.

Do you like this approach? Should I extend it to all the APIs?

michaelherger: I'd use constants whenever there's static string literals to compare otherwise. If you know of other places where this happens I'd recommend you change that. I only commented on the changes in this PR...

darrell-k: I'll check for string literals.

What I was trying to say is that the APIs use a mixture of positional parameters and named parameters in a hash. I've changed this for the favorite* APIs so that named parameters are always used. 

I was wondering if I should apply the same to the rest of the APIs?

SamInPgh:
I've been running with these changes for a couple of weeks now and have noticed no behavioral changes. Other than that, I have nothing to add review-wise. Is this ready to publish?

SamInPgh (approved):
Looks good to me.

darrell-k:
@SamInPgh have you been using the version with the most recent commit https://github.com/LMS-Community/plugin-Qobuz/pull/127/changes/6cd918492884500e4cd3ad52dcfecffdef8cc13e ? If so, yes let's merge.

SamInPgh:
> @SamInPgh have you been using the version with the most recent commit [6cd9184](https://github.com/LMS-Community/plugin-Qobuz/commit/6cd918492884500e4cd3ad52dcfecffdef8cc13e) ? If so, yes let's merge.

Yes, I have. Merging...

PR #128, opened 2026-04-04
Miscellaneous replay gain improvements.

SamInPgh:
1. Use player's "Default Adjustment for Remote Streams" prefs value for track gain if "replay_gain" tag is undefined for the track. It's rare but does occur.
2. Remove now unnecessary check for first track on an album when calculating "Smart Gain".
3. Improve readability of Smart Gain Info logging messages.


michaelherger (approved):
👍🏻

PR #132, opened 2026-08-21
Disable "MyWeeklyQ", etc.

SamInPgh:
1. Disable "MyWeeklyQ".
2. Populate the `bitrate` attribute with estimated values based on sample rate and depth for tracks imported into the library.

I wanted these commits to be two separate PR's, but Github had it's own ideas. 

P.S. Take it easy on me in the review. It's my birthday.

### DIFF
--- a/Importer.pm
+++ b/Importer.pm
@@ -335,6 +335,7 @@ sub _prepareTrack {
 
 	my $url = Plugins::Qobuz::API::Common->getUrl(undef, $track) || return;
 	my $ct  = Slim::Music::Info::typeFromPath($url);
+	my $bitrate = Slim::Player::Song::guessBitrateFromFormat($ct, undef, $track->{maximum_bit_depth}, $track->{maximum_sampling_rate} * 1000);
 
 	my $attributes = {
 		url          => $url,
@@ -354,6 +355,7 @@ sub _prepareTrack {
 		SAMPLESIZE   => $track->{maximum_bit_depth},
 		CHANNELS     => $track->{maximum_channel_count},
 		LOSSLESS     => $ct eq 'flc',
+		BITRATE      => $bitrate,
 		RELEASETYPE  => $album->{release_type} =~ /^[a-z]+$/ ? ucfirst($album->{release_type}) : $album->{release_type},
 		REPLAYGAIN_ALBUM_GAIN => $album->{replay_gain},
 		REPLAYGAIN_ALBUM_PEAK => $album->{replay_peak},
--- a/Importer.pm
+++ b/Importer.pm
@@ -335,7 +335,7 @@ sub _prepareTrack {
 
 	my $url = Plugins::Qobuz::API::Common->getUrl(undef, $track) || return;
 	my $ct  = Slim::Music::Info::typeFromPath($url);
-	my $bitrate = Slim::Player::Song::guessBitrateFromFormat($ct, undef, $track->{maximum_bit_depth}, $track->{maximum_sampling_rate} * 1000);
+	my $bitrate = ($ct eq 'mp3' ? 320 : $track->{maximum_bit_depth} * $track->{maximum_sampling_rate}) * 1000;
 
 	my $attributes = {
 		url          => $url,
--- a/Plugin.pm
+++ b/Plugin.pm
@@ -365,11 +365,11 @@ sub handleFeed {
 		image => 'html/images/genres.png',
 		type => 'link',
 		url  => \&QobuzGenres
-	},{
-		name => cstring($client, 'PLUGIN_QOBUZ_MYWEEKLYQ'),
-		type  => 'playlist',
-		url  => \&QobuzMyWeeklyQ,
-		image => 'html/images/playlists.png'
+#	},{
+#		name => cstring($client, 'PLUGIN_QOBUZ_MYWEEKLYQ'),
+#		type  => 'playlist',
+#		url  => \&QobuzMyWeeklyQ,
+#		image => 'html/images/playlists.png'
 	}];
 
 	if ($client && scalar @{ Plugins::Qobuz::API::Common::getAccountList() } > 1) {


Importer.pm
```diff
@@ -335,6 +335,7 @@ sub _prepareTrack {
 
 	my $url = Plugins::Qobuz::API::Common->getUrl(undef, $track) || return;
 	my $ct  = Slim::Music::Info::typeFromPath($url);
+	my $bitrate = ($ct eq 'mp3' ? 320 : $track->{maximum_bit_depth} * $track->{maximum_sampling_rate}) * 1000;
```
darrell-k, line 338: Would we need to multiply by the number of channels?

Plugin.pm
```diff
@@ -365,11 +365,11 @@ sub handleFeed {
...
 		url  => \&QobuzGenres
-	},{
-		name => cstring($client, 'PLUGIN_QOBUZ_MYWEEKLYQ'),
-		type  => 'playlist',
-		url  => \&QobuzMyWeeklyQ,
-		image => 'html/images/playlists.png'
+#	},{
+#		name => cstring($client, 'PLUGIN_QOBUZ_MYWEEKLYQ'),
+#		type  => 'playlist',
+#		url  => \&QobuzMyWeeklyQ,
```
darrell-k, line 371: Get rid of the subroutine, too?

darrell-k:
Happy birthday!

SamInPgh:
> Happy birthday!

Thanks!

SamInPgh:
@darrell-k  Do you have any comments or suggestions regarding these changes? If not, I was going to push a new release.

darrell-k:
Did you see my two review comments, above?

SamInPgh:
> Did you see my two review comments, above?

I only see the "Happy birthday" one. 🤔

darrell-k:
Oops, forgot to press the "submit review" button. You should see them now.

Importer.pm
SamInPgh: No. The result of not doing so, as stated in the comments of `guessBitrateFromFormat()`, is to estimate a 50% compression rate (which makes sense as long as it is a 2-channel track). Of course it would be more accurate to multiply by the number of channels and then divide by 2 but I'm fairly certain that all of the Qobuz tracks are of the 2-channel variety.

Plugin.pm
SamInPgh: The intent was to disable the functionality while leaving the backing logic in place for possible future use if the API should be restored. Is there some harm I have overlooked in leaving the orphaned subroutine intact, or do you feel that it would be better to remove all traces of the feature?

Importer.pm
darrell-k: Depending on the nature of the signal and the flac compression level used, the compression might be way higher than 50%. But more importantly, it's unpacked on playback so what we actually hear is the original bit rate.

So my view is that the flac-compressed bit rate is a meaningless number, even if we calculated it accurately (which we could do using the track duration and file size).

The bit depth and sample rate define the quality exactly, bitrate doesn't matter like it does for lossy formats. It's just another way of expressing bit depth/sample rate.

I believe showing it this way will only lead to confusion, especially amongst people who don't understand lossless compression and will think the compressed bit rate is somehow an indication of lower quality.

Plugin.pm
darrell-k: I don't think it really matters. It's easily enough restored in the future from git history, but if you decide to leave it, maybe a comment saying not currently in use?

Importer.pm
SamInPgh: > Depending on the nature of the signal and the flac compression level used, the compression might be way higher than 50%. 

True. It's only an estimate.

>But more importantly, it's unpacked on playback so what we actually hear is the original bit rate.
> 

Obviously. What's your point though? And why is it so important? I'm lost.

> So my view is that the flac-compressed bit rate is a meaningless number, even if we calculated it accurately (which we could do using the track duration and file size).
> 

It's not meaningless at all. In fact, it's required in order to support jumping backward or forward in a track based on time offset, which I discovered awhile ago while fixing that support for Qobuz tracks. We DO calculate it accurately during playback because we have access to the file size from the flac header. We don't have access to the file size during the import. However the estimate provides, in my opinion, a useful placeholder for it in the library until playback is started --- certainly more useful than a constant '705kbps' for every track that is currently displayed when no value is present.

> The bit depth and sample rate define the quality exactly, bitrate doesn't matter like it does for lossy formats. It's just another way of expressing bit depth/sample rate.
> 

Why are we debating the existential value of bitrates for lossless tracks here? I'm confused. Isn't that a different discussion?

> I believe showing it this way will only lead to confusion, especially amongst people who don't understand lossless compression and will think the compressed bit rate is somehow an indication of lower quality.

What can I say? I completely disagree. How about if we leave it up to those easily confused and uneducated users to express their ignorance? I will personally answer all of their poorly worded questions. Seriously, though. I'm the one that's confused right now. I wasn't expecting this reaction. It seems like more of a philosophical treatise than a code review.

darrell-k: Of course the bit rate is important for seeking +/- within a track. I just meant as a database column value. 

My view is that in the database, if we need to store it, it should be the playback bit rate, as determined by the bit depth/sample rate. After all, that is what the user actually hears. 

I'm sorry that you feel I've gone beyond a code review. Maybe I should have first asked what is this for? ie what is the stored value to be used for? If it is simply to show to the user as a track attribute, my feeling is that it should be the actual bit rate they will experience when they play the track, eg 1,411,200 or 4,608,000 or whatever.

I don't think that's philosophical, it's about functionality.

darrell-k: @SamInPgh I think I owe you an apology: I've just looked at LMS and for local flac tracks it calculates a bitrate from the file size/track length. So I can see now that what you are doing is trying to replicate that as closely as possible without the required information.

So I'd say go ahead with this change as is.

However, I do believe that for lossless tracks LMS should not set the bit rate this way, rather as I say above store the playback bit rate for lossless files. My feeling is that the current bitrate calculation approach is a hangover from the lossy file days. I may prepare an LMS PR to change this.

I don't know if such a PR would meet approval, but if it did I would also add code to the online import module in LMS so that all online imports followed the same logic for lossless files, avoiding having to change plugins individually at that stage.

As an aside, looking at my own library, for 16/44.1 stereo flac rips, the bitrate written to the tracks table varies between 375,000 and 1,100,000! The lower end seems to be single instrument stuff and the higher end is mostly live rock recordings.

SamInPgh: > Of course the bit rate is important for seeking +/- within a track. I just meant as a database column value.
> 

Okay. So you're questioning the existential value of the 'bitrate' database column in general for all tracks, not just those imported from Qobuz, the difference being that the value for local tracks can be calculated accurately by dividing the file size by the duration. That's what I meant by it becoming a philosophical debate.

> My view is that in the database, if we need to store it, it should be the playback bit rate, as determined by the bit depth/sample rate. After all, that is what the user actually hears.
> 

The user doesn't hear bits. They hear the audio signal represented by a stream of them that are sent to the player by LMS in a compressed bitstream (for flac) at the average rate indicated in the 'bitrate' attribute. (Let's leave transcoding out of the discussion for now.) The bitstream is then uncompressed by the player hardware/software back to it's original size and converted to analog audio. You seem to be advocating that the 'bitrate' value stored in the database for all tracks, including flac, should be simply `samplesize*samplerate*channels`, which would indeed render its existence unnecessary. However, that value doesn't represent the rate at which bits are sent to the player by LMS. Rather, it represents the rate at which the bits are processed by the DAC after being decompressed for conversion to analog. Does that make sense? Please correct me if I'm mistaken. If I'm not, then it comes down to a question of "What is the meaning of 'bitrate', as used in describing a lossless compressed digital audio signal and does it have any intrinsic value as a technical specification?", which sure sounds like a philosophical question to me. 😉

> I'm sorry that you feel I've gone beyond a code review. Maybe I should have first asked what is this for? ie what is the stored value to be used for? If it is simply to show to the user as a track attribute, my feeling is that it should be the actual bit rate they will experience when they play the track, eg 1,411,200 or 4,608,000 or whatever.
> 

But again, this would apply to all flac tracks stored in the library, including local ones. Right?

> I don't think that's philosophical, it's about functionality.

See my comment above. 👆

darrell-k: See my subsequent post.

SamInPgh: No apology necessary. I must admit, however, that I was beginning to consider the possibility that my heretofore slow but steady descent into senility had begun to quicken it's pace as I enter my 78th year. 🤤

